PR #66, opened 2018-09-27
Add MemoryLow to service file

ojmhetar:


ojmhetar:
While I have not seen a warning message, I've noticed when using the legacy hierarchy, the ``MemoryLow`` setting is acknowledged on a ``systemctl status`` . 

image

As far as I know, the easiest way to determine the hierarchy is to check the path ``/sys/fs/cgroup`` (or wherever cgroup is mounted). cgroupv1 has a hierarchy per-resource so the resources are shown as directories: cpu/, memory/, pids/, blkio/, etc. In cgroupv2, you would see the cgroup themselves as directories: system.slice, user.slice, kubepods, etc.

dlipovetsky:
@ojmhetar Thanks! I see that Container Linux is using cgroupv2, since I see directories like `/sys/fs/cgroup/memory/kubepods/podaa000653-c358-11e8-b0e4-fa163e85d5ff/`

PR #79, opened 2018-12-18
Rerun dep

justinsb:
It looks like vendor had previously been run with a prior version of
dep.


dlipovetsky:
@justinsb I merged #80 before this by mistake. But maybe this PR is not needed--running `dep ensure` (with dep 0.5.0) on master results in no changes.

dlipovetsky:
Closing, since go modules replaced dep in #103

PR #94, opened 2019-03-11
fix #77: add `--listen-address` and --advertise-address`

### LINKED ISSUE #77: Add flag to specify listening address.
This is necessary for environments like Vagrant that have multiple network interfaces.

if I could specify the ip that is used to resolve the external address this would solve this.

spiette:
This commit add two options to configure etcd to bind on a specific address or to advertise a specific address, locally assigned or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,11 @@ Etcdadm must be run on the machine that is being added or removed. As a conseque
 On its own, etcdadm does not automate cluster operation, but a cluster orchestrator can delegate all the above tasks to etcdadm.
 
 <a name="questions"></a>
+
+## Options
+
+*  `--listen-address`, `--advertise-address`: specify an IP if you'd like something different than the auto-discovery. Regardless of this, you can also specify an alternative advertised address (for a floating IP for example).
+
 ## Questions?
 
 For more information reach out to [etcdadm slack channel](https://kubernetes.slack.com/messages/CEM0AT9GW)
--- a/apis/config.go
+++ b/apis/config.go
@@ -69,6 +69,9 @@ type EtcdAdmConfig struct {
 
 	LoopbackClientURL url.URL
 
+	ListenAddress       string
+	AdvertiseAddress    string
+
 	// ServerCertSANs sets extra Subject Alternative Names for the etcd server signing cert.
 	ServerCertSANs []string
 	// PeerCertSANs sets extra Subject Alternative Names for the etcd peer signing cert.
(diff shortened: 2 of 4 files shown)


apis/config.go
```diff
@@ -288,3 +305,19 @@ func defaultExternalAddress() (net.IP, error) {
 	}
 	return ip, nil
 }
+
+func defaultListenAddress(cfg *EtcdAdmConfig) (net.IP, error) {
+	if cfg.ListenAddress == "" {
+		return defaultExternalAddress()
+	} else {
+		return netutil.ChooseBindAddress(net.ParseIP(cfg.ListenAddress))
```
dlipovetsky: `--listen-address 127.0.0.1` is a no-op, because of `ChooseBindAddress`. I think users should be allowed to use `127.0.0.1` as a listen address.

What about implementing `defaultExternalAddress` as in [kube-apiserver](https://github.com/kubernetes/kubernetes/blob/a66c6e459cad47ce39eb3a6d3f6db380ac1db21c/staging/src/k8s.io/apiserver/pkg/server/options/serving.go#L111)?

Then `defaultListenAddress` and `defaultAdvertiseAddress` can just wrap that and add nice context to errors.

apis/config.go
```diff
@@ -263,13 +273,20 @@ func DefaultLoopbackClientURL(cfg *EtcdAdmConfig) {
...
-	cfg.ListenClientURLs = cfg.AdvertiseClientURLs
+	externalAddress, err := defaultListenAddress(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to set default ListenClientURLs: %s", err)
+	}
+	cfg.ListenClientURLs = append(cfg.ListenClientURLs, url.URL{
+		Scheme: "https",
+		Host: fmt.Sprintf("%s:%d", externalAddress.String(), constants.DefaultClientPort),
+	})
 	cfg.ListenClientURLs = append(cfg.ListenClientURLs, cfg.LoopbackClientURL)
```
dlipovetsky: Any thoughts on the fact that the `LoopbackClientURL` is always used?

README.md
```diff
@@ -79,6 +79,11 @@ Etcdadm must be run on the machine that is being added or removed. As a conseque
 On its own, etcdadm does not automate cluster operation, but a cluster orchestrator can delegate all the above tasks to etcdadm.
 
 <a name="questions"></a>
+
+## Options
+
+*  `--listen-address`, `--advertise-address`: specify an IP if you'd like something different than the auto-discovery. Regardless of this, you can also specify an alternative advertised address (for a floating IP for example).
```
dlipovetsky: This is fine, although it conflates peer and client addresses. In the future, we should allow users to configure peer and client addresses separately.

Can you note that the address is used for derive both peer and client URLs?

README.md
```diff
@@ -79,6 +79,11 @@ Etcdadm must be run on the machine that is being added or removed. As a conseque
...
+
 ## Questions?
 
-For more information reach out to [etcdadm slack channel](https://kubernetes.slack.com/messages/CEM0AT9GW)
\ No newline at end of file
+For more information reach out to [etcdadm slack channel](https://kubernetes.slack.com/messages/CEM0AT9GW)
```
dlipovetsky: (Note to reviewers: this adds a newline)

dlipovetsky:
@spiette Are you able to address the review comments?

PR #95, opened 2019-03-22
cmd: print artifact download url

nirmoy:
etcdadm printing err which nil  instead of etcdAdmConfig.ReleaseURL, this PR fix that.


dlipovetsky:
@nirmoy Thanks for the fix!

/approve

PR #104, opened 2019-05-16
Add sequence diagrams for init, join, and reset

dlipovetsky:
Images generated by sequencediagram.org


docs/diagrams/join.md
```diff
@@ -0,0 +1,41 @@
...
+group download (if not in cache)
+etcdadm->remote repo: download etcd, etcdctl into local cache
+end
+
+group install artifacts
+etcdadm->etcdadm: copy etcd, etcdctl from local cache to install directory
+end
+
+group create certificates
+etcdadm->etcdadm: create CA certificate
```
astrieanna: Join doesn't create a CA cert, right? It uses the CA cert passed in by the user, the one that was created in init?

docs/diagrams/reset.md
```diff
@@ -0,0 +1,32 @@
+title etcdadm join https://member1:2379
```
astrieanna: I think this label is wrong (should be reset)

docs/diagrams/join.md
```diff
@@ -0,0 +1,41 @@
+title etcdadm join https://member1:2379
+
+participant user
+participant remote repo
+participant etcdadm
+participant systemd
+database member1
+database member2
```
astrieanna: Because the sequence diagram is showing the interaction between multiple processes on one host (etcdadm, systemd, member2) and also between hosts, it feels a bit confusing. It looks like the website you're using has a `participantgroup` concept. Do you think grouping the process participants by host would help?

docs/diagrams/init.md
```diff
@@ -0,0 +1,32 @@
...
+database member1
+
+user->etcdadm: init
+
+group download (if not in cache)
+etcdadm->remote repo: download etcd, etcdctl into local cache
+end
+
+group install artifacts
+etcdadm->host1: copy etcd, etcdctl from local cache to install directory
```
astrieanna: Two of the participants in this diagram (`etcadm`, `systemd`) are processes; `host1` is not. In order to make the participants more parallel/similar, maybe `host1` here should be a `database` participant or be called the `filesystem of host1`.

docs/diagrams/reset.md
```diff
@@ -0,0 +1,32 @@
...
+etcdadm->member1: remove member2
+destroy member2
+end
+
+group uninstall artifacts
+etcdadm->etcdadm: delete etcd, etcdctl from install directory
+end
+
+group destroy certificates
+etcdadm->etcdadm: destroy CA certificate
```
astrieanna: If `etcdadm reset` deletes the CA cert, then if you want to run `etcdadm join` again on the same machine (to the same cluster), you'll need to copy the CA cert over again, right?

docs/diagrams/join.md
dlipovetsky: Right. Join doesn't create a CA cert. Thanks for catching this!

docs/diagrams/reset.md
dlipovetsky: You're right. Copy-paste error. Thanks!

docs/diagrams/reset.md
dlipovetsky: Yes, that's what currently happens.

docs/diagrams/join.md
dlipovetsky: Yes! Great idea. Will add this.

docs/diagrams/init.md
dlipovetsky: Good idea. I'll use `filesystem` in the `host1` `participantgroup`.

dlipovetsky:
@astrieanna Thanks a lot for your review. I tried to address your comments.

justinsb:
Thanks @dlipovetsky - looks great.  One suggestion would be to include a note in the repo about how to generate the pngs

astrieanna:
It looks like in addressing the review comments, the image files were not updated.

dlipovetsky:
@astrieanna My fault. I updated the image files. I also changed to PlantUML so they can be re-generated at the command line.

@justinsb Good point. I added a make target to generate diagrams.

astrieanna:
These diagrams look great. :)

PR #105, opened 2019-06-02
README: add command example of copying ca cert and key

haoqing0110:
etcdadm is awesome, and this PR help people better knowing what and where to copy the cert and key.


dlipovetsky:
Thanks for adding this!

/lgtm
/approve

PR #113, opened 2019-07-04
only download etcd archive if it does not exist

ghouscht:
Simple check to download the etcd archive only if it does not exist alredy - fixes issue #73

### DIFF
--- a/binary/etcd.go
+++ b/binary/etcd.go
@@ -107,13 +107,20 @@ func extract(extractDir, archive string) error {
 	return nil
 }
 
-// Download installs the etcd binaries in the directory specified by locationDir
+// Download downloads the etcd binaries in the directory specified by locationDir
 func Download(releaseURL, version, locationDir string) error {
+	archive := filepath.Join(locationDir, releaseFile(version))
+	// if desired version is already downloaded, do nothing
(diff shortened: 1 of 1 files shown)


dlipovetsky:
Thanks @ghouscht!

I want to note this introduces a new issue: if the archive is corrupt or incomplete, etcdadm will not download it again, and the user needs to delete it. We can address that issue by verifying the archive using the published key. I'll file a separate PR.

/approve
/lgtm

Let's wait a couple of days for comments. @moshloop and @justinsb .
/hold

ghouscht:
> I want to note this introduces a new issue: if the archive is corrupt or incomplete, etcdadm will not download it again, and the user needs to delete it. We can address that issue by verifying the archive using the published key. I'll file a separate PR.

Good note! This should be done anyway. I could have a look at it in the next few days and adopt my PR if this is ok.

dlipovetsky:
@ghouscht I think we can merge this and address my comment separately. Thanks for this 

/hold cancel

dlipovetsky:
> docs/USECASES.md: no newline at the end of file

Sorry, an unrelated error causes the test job to fail. Fixing momentarily.

PR #129, opened 2019-09-26
bugfix: FileMode should be octal expressed

beautytiger:
Add prefix `0`,  or it will be translated to decimal.


dlipovetsky:
Thanks for the fix @beautytiger!

/lgtm
/approve

PR #131, opened 2019-09-26
remove vendor

chuckha:
Using go.mod vendor is no longer required. There was one minor change to the go.sum file.

Signed-off-by: Chuck Ha <chuckh@vmware.com>


dlipovetsky:
@chuckha  Is it correct to say that the `vendor` directory is already being ignored if using go 1.12, so this change just removes that directory?

chuckha:
if this is outside the go path then by default, yes, go will ignore vendor

Issue #124, opened 2019-09-16
Logs are a bit confusing by default


chuckha:
I just ran `etcdadm init` and got this output:

```
INFO[0000] [install] Artifact not found in cache. Trying to fetch from upstream: https://github.com/coreos/etcd/releases/download 
INFO[0000] [install] Downloading & installing etcd https://github.com/coreos/etcd/releases/download from 3.3.8 to /var/cache/etcdadm/etcd/v3.3.8 
INFO[0000] [install] downloading etcd from https://github.com/coreos/etcd/releases/download/v3.3.8/etcd-v3.3.8-linux-amd64.tar.gz to /var/cache/etcdadm/etcd/v3.3.8/etcd-v3.3.8-linux-amd64.tar.gz 
######################################################################## 100.0%
INFO[0002] [install] extracting etcd archive /var/cache/etcdadm/etcd/v3.3.8/etcd-v3.3.8-linux-amd64.tar.gz to /tmp/etcd242083477 
INFO[0002] [install] verifying etcd 3.3.8 is installed in /opt/bin/ 
INFO[0002] [certificates] creating PKI assets           
INFO[0002] creating a self signed etcd CA certificate and key files 
[certificates] Generated ca certificate and key.
INFO[0003] creating a new server certificate and key files for etcd 
[certificates] Generated server certificate and key.
[certificates] server serving cert is signed for DNS names [ip-10-0-0-19] and IPs [10.0.0.19 127.0.0.1]
INFO[0003] creating a new certificate and key files for etcd peering 
[certificates] Generated peer certificate and key.
[certificates] peer serving cert is signed for DNS names [ip-10-0-0-19] and IPs [10.0.0.19]
INFO[0003] creating a new client certificate for the etcdctl 
[certificates] Generated etcdctl-etcd-client certificate and key.
INFO[0003] creating a new client certificate for the apiserver calling etcd 
[certificates] Generated apiserver-etcd-client certificate and key.
[certificates] valid certificates and keys now exist in "/etc/etcd/pki"
INFO[0005] [health] Checking local etcd endpoint health 
INFO[0005] [health] Local etcd endpoint is healthy      
INFO[0005] To add another member to the cluster, copy the CA cert/key to its certificate dir and run: 
INFO[0005] 	etcdadm join https://10.0.0.19:2379         
```

It looks like a mix of `klog` and `fmt`. It'd be nice if it were one or the other

/priority longterm-important

[...]

PR #132, opened 2019-10-08
Logging consolidation

frederiko:
Initial PR to standardize most logs to use `logrus` in replacement for pure `fmt` (except for `Errorf()`). (#124)
Also, in certain cases, added the component/stage (e.g. [certificates]) as proposed

### DIFF
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -53,7 +53,7 @@ func CreatePKIAssets(cfg *apis.EtcdAdmConfig) error {
 		}
 	}
 
-	log.Printf("[certificates] valid certificates and keys now exist in %q\n", cfg.CertificatesDir)
+	log.Printf("[certificates] Valid certificates and keys now exist in %q\n", cfg.CertificatesDir)
 
 	return nil
 }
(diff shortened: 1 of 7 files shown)


frederiko:
I signed it

dlipovetsky:
Thanks @frederiko!

The reason for using both `fmt` and `log` was to use both stdout and stderr, respectively. 

When I read #124, I think of the message content and structure (e.g. timestamps, prefixes). I think that can be improved, but I think we should continue to use both stdout and stderr.

WDYT?

/cc @chuckha @CamelCaseNotation

frederiko:
Hi @dlipovetsky 
I think it makes sense and I believe I have the same understanding. The piece I tried to initially address here was this comment: "It looks like a mix of klog and fmt. It'd be nice if it were one or the other". Perhaps I got it wrong.

frederiko:
"I signed it"

neolit123:
klog makes sense for `klog.Warning` we should be shown at all times (to stderr) and `klog.V(x).` which should be enabled for higher verbosity.

ideally klog output without verbosity should be mixed with `fmt.Print*` calls (to stdout).

at least this is what we went for in kubeadm.

frederiko:
I made a mistake and didn't realize that `klog` != `logrus`. I took `logrus` as it was already in use.
Is it a fair assumption that it should be continue using `logrus` just as proposed here? 

And, what do you guys recommend here?

neolit123:
ideally it should be either logrus or klog, the majority of k8s projects use klog.

frederiko:
Well, in that case, if there are not objections, I could help in getting it migrated to klog.

neolit123:
+1 on using an interface in all places instead of binding an external logger as a dependency in all files.

dlipovetsky:
Thanks for your feedback @neolit123 and @chuckha. And thanks for your patience @frederiko! :smile:

Just to recap: As I understood  it, changing `fmt` to `log` will change the destination from stdout to stderr. I would like to avoid that, because, in my experience, both stdout and stderr are useful, e.g., to write progress to stdout, and to write warnings/errors to stderr.

I agree that, because the messages are formatted differently, it is hard to read them interleaved. I would support using the same format everywhere, but continuing to write some messages to stdout and others to stderr. 

I have no opinon on the implementation details. If klog does what we need, I think that's fine.

PR #139, opened 2019-11-07
fix the build of misspell for go 1.13

neolit123:
this was overlooked for go 1.12 and broke when we started using 1.13.

/kind bug
/priority important-soon
/assign @dlipovetsky


dlipovetsky:
Thanks @neolit123!

/lgtm

Issue #140, opened 2019-11-29
Cannot find package running go get


perezjasonr:
Hello, I'm wondering if there is some kind of dependency mixup and/or if theres a disconnect with what running a build for this is expecting. Does something need to be updated?

When running a simple:

`go get -u github.com/kubernetes-sigs/etcdadm`
or
`go get -u sigs.k8s.io/etcdadm`

I get:

```
package k8s.io/kubernetes/pkg/version: cannot find package "k8s.io/kubernetes/pkg/version" in any of:
        /usr/local/go/src/k8s.io/kubernetes/pkg/version (from $GOROOT)
        /home/centos/go/src/k8s.io/kubernetes/pkg/version (from $GOPATH)
package github.com/coreos/etcd/snapshot: cannot find package "github.com/coreos/etcd/snapshot" in any of:
        /usr/local/go/src/github.com/coreos/etcd/snapshot (from $GOROOT)
        /home/centos/go/src/github.com/coreos/etcd/snapshot (from $GOPATH)
```

There certainly is no "version" under go/src/k8s.io/kubernetes/pkg:

```
sudo ls -la /home/centos/go/src/k8s.io/kubernetes/pkg/
total 56
drwxrwxr-x. 33 centos centos 4096 Nov 29 15:14 .
drwxrwxr-x. 19 centos centos 4096 Nov 29 15:14 ..
drwxrwxr-x. 12 centos centos  201 Nov 29 15:14 api
drwxrwxr-x. 26 centos centos 4096 Nov 29 15:14 apis
drwxrwxr-x.  4 centos centos   60 Nov 29 15:14 auth
-rw-rw-r--.  1 centos centos 3124 Nov 29 15:14 BUILD
drwxrwxr-x.  2 centos centos   84 Nov 29 15:14 capabilities
drwxrwxr-x.  6 centos centos   95 Nov 29 15:14 client
drwxrwxr-x.  3 centos centos   81 Nov 29 15:14 cloudprovider
drwxrwxr-x. 33 centos centos 4096 Nov 29 15:14 controller
drwxrwxr-x.  6 centos centos  238 Nov 29 15:14 credentialprovider
drwxrwxr-x.  2 centos centos   57 Nov 29 15:14 features
drwxrwxr-x.  2 centos centos   78 Nov 29 15:14 fieldpath
drwxrwxr-x.  3 centos centos   48 Nov 29 15:14 generated
[...]

PR #141, opened 2019-12-05
decouple etcdadm from kubernetes/kubernetes

neolit123:
To be able to decouple etcdadm from kubernetes/kubernetes the following
changes were required:

- use k/component-base/version instead of k/k/pkg/version
- sync the pki_helpers fork
- update other module dependencies
- use the new ResolveBindAddress() function from apimachinery/utils/net

potential fix for https://github.com/kubernetes-sigs/etcdadm/issues/140

/kind cleanup
/priority important-longterm
/assign @dlipovetsky


dlipovetsky:
Thank you @neolit123, for decoupling and syncing with upstream!

/approve
/lgtm 

I will test `go get` once this merges.

PR #143, opened 2019-12-16
Ignore vendor directory when git push

pytimer:
I update the lasted code, use `go mode vendor` that save the third party library to `vendor` directory, I found the `.gitignore` not ignore the `vendor` directory, so add this entry in `.gitignore` file.


dlipovetsky:
Thanks!

/ok-to-test

Issue #144, opened 2019-12-16
Download etcd release package from custom url error


pytimer:
I use the lasted etcdadm to init the cluster, want to use `--release-url`  download etcd release package from my custom url, i save the etcd release package to wrong path, but i see `etcdadm` also download completed, the error occur to extracting archive package.

error logs:

```sh
[root@k8s-1 ~]# etcdadm init --name k8s --release-url http://192.168.10.108/tools/etcd --version 3.3.18 -l debug
INFO[0000] [install] Artifact not found in cache. Trying to fetch from upstream: http://192.168.10.108/tools/etcd
INFO[0000] [install] Downloading & installing etcd http://192.168.10.108/tools/etcd from 3.3.18 to /var/cache/etcdadm/etcd/v3.3.18
INFO[0000] [install] downloading etcd from http://192.168.10.108/tools/etcd/v3.3.18/etcd-v3.3.18-linux-amd64.tar.gz to /var/cache/etcdadm/etcd/v3.3.18/etcd-v3.3.18-linux-amd64.tar.gz
######################################################################## 100.0%
INFO[0000] [install] extracting etcd archive /var/cache/etcdadm/etcd/v3.3.18/etcd-v3.3.18-linux-amd64.tar.gz to /tmp/etcd862140673

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
FATA[0000] [install] Artifact could not be installed from cache: unable to extract etcd archive: exit status 2
```

Can `etcdadm` downloading etcd from `--release-url` show error instead of extract etcd archive if the package not found ?

dlipovetsky:
> Can etcdadm downloading etcd from --release-url show error instead of extract etcd archive if the package not found ?

@pytimer, are you saying that the curl call ignores errors? Because etcdadm [should fail if curl reports an error](https://github.com/kubernetes-sigs/etcdadm/blob/c264b49a95a4e4f5cb7211fa424546bdeacdd72c/binary/etcd.go#L87).

pytimer:
@dlipovetsky Yes.I think etcdadm call curl should not ignore errors, so i add `--fail` when invoke curl command to download etcd release package.

dlipovetsky:
> @dlipovetsky Yes.I think etcdadm call curl should not ignore errors,

I agree with you. Unfortunately, I couldn't understand the problem from your report, but I see it now; curl will exit with zero if it receives some output with a 404.

```
[...]

PR #145, opened 2019-12-16
Download etcd release abort and show error if package not found

pytimer:
Fixes: https://github.com/kubernetes-sigs/etcdadm/issues/144

Add `--fail` when using `curl` download etcd release package. 

If etcd release package not found, `etcdadm` will show below logs:

```diff
[root@k8s-1 ~]# etcdadm init --name k8s --release-url http://192.168.10.108/tools/etcd --version 3.3.12 -l debug
INFO[0000] [install] Artifact not found in cache. Trying to fetch from upstream: http://192.168.10.108/tools/etcd
INFO[0000] [install] Downloading & installing etcd http://192.168.10.108/tools/etcd from 3.3.12 to /var/cache/etcdadm/etcd/v3.3.12
INFO[0000] [install] downloading etcd from http://192.168.10.108/tools/etcd/v3.3.12/etcd-v3.3.12-linux-amd64.tar.gz to /var/cache/etcdadm/etcd/v3.3.12/etcd-v3.3.12-linux-amd64.tar.gz

+curl: (22) The requested URL returned error: 404 Not Found
FATA[0000] [install] Unable to fetch artifact from upstream: unable to download etcd: exit status 22
```

### DIFF
--- a/binary/etcd.go
+++ b/binary/etcd.go
@@ -76,7 +76,7 @@ func isEtcdctlInstalled(version, inputDir string) (bool, error) {
 
 func get(url, archive string) error {
 	log.Printf("[install] downloading etcd from %s to %s\n", url, archive)
-	argStr := fmt.Sprintf("--connect-timeout %v --progress-bar --location --output %v %v", int(constants.DefaultDownloadConnectTimeout/time.Second), archive, url)
+	argStr := fmt.Sprintf("--fail --connect-timeout %v --progress-bar --location --output %v %v", int(constants.DefaultDownloadConnectTimeout/time.Second), archive, url)
 	cmd := exec.Command("curl", strings.Fields(argStr)...)
 	cmd.Stdout = os.Stdout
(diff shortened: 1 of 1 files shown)


chuckha:
/lgtm

Definitely makes sense to fail on 404s. Copying in the man page to show why `--fail` is good for scripts and the one warning about using it:

```
       -f, --fail
              (HTTP)  Fail  silently  (no output at all) on server errors. This is mostly done to better enable scripts etc to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will prevent curl from outputting
              that and return error 22.

              This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407).
```

dlipovetsky:
/approve

Thanks @pytimer for finding and fixing the issue, and @chuckha for the review!

Issue #133, opened 2019-10-16
Joining a third member fails (sometimes)


chuckha:
There appears to be a health check of the cluster when joining a node. If you join two nodes in quick succession the second one will fail since the cluster takes a second to get healthy for a third node to join

I recommend adding a retry with backoffs of 1, 3, 5 seconds which would cap the total wait time to 9 seconds for joining a node.

/kind bug

pytimer:
@chuckha I add exponential backoff retry for join member, i already testing add two members at the same time  in my environment, it works. You can review and test it if you have time.

PR #146, opened 2019-12-17
Add backoff retry when join member at the same time

pytimer:
Fixes: https://github.com/kubernetes-sigs/etcdadm/issues/133

Add exponential backoff retry for join etcd member at the same time, e.g. 0, 1, 2, 4, 8s. The max retry is five times.

### DIFF
--- a/apis/config.go
+++ b/apis/config.go
@@ -85,6 +85,10 @@ type EtcdAdmConfig struct {
 
 	Snapshot      string
 	SkipHashCheck bool
+
+	// Retry sets enable or disable backoff retry when join etcd member to cluster.
+	// Default true, it mean that enable backoff retry.
+	Retry bool
 }
 
 // EndpointStatus TODO: add description
--- a/cmd/join.go
+++ b/cmd/join.go
@@ -35,6 +35,7 @@ import (
 	"sigs.k8s.io/etcdadm/service"
 
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var joinCmd = &cobra.Command{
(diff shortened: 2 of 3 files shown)


dlipovetsky:
/ok-to-test

Thanks @pytimer! I'll try to review this soon.

chuckha:
Sweet! this worked really well!
/lgtm

pytimer:
@chuckha I have tried many times join member, basically the waiting time is 4s, so should we increase every retry backoff duration?

chuckha:
> @chuckha I have tried many times join member, basically the waiting time is 4s, so should we increase every retry backoff duration?

Instead of 1, 1, 2, 2, 4, etc? I think it works fine as is, I'm not sure what you're suggesting here really

dlipovetsky:
> @chuckha I have tried many times join member, basically the waiting time is 4s, so should we increase every retry backoff duration?

Given these parameters:
```
	DefaultBackOffSteps    = 5
	DefaultBackOffDuration = 1 * time.Second
	DefaultBackOffFactor   = 2.0
```
The maximum duration is 10 seconds, right? The parameters seem fine.

dlipovetsky:
I apologize for my late review! I would prefer making retry/backoff optional using a flag. Sometimes, retrying is not desirable. WDYT?

lgtm otherwise

pytimer:
> Instead of 1, 1, 2, 2, 4, etc? I think it works fine as is, I'm not sure what you're suggesting here really

@chuckha Sorrry, i mean join two member at the same time, `etcdadm` backoff retry three times basically, it takes about 4s， so i want to update `DefaultBackOffDuration = 2 * time.Second`, and backoff time is 0, 2, 4, 8, 16.

> I apologize for my late review! I would prefer making retry/backoff optional using a flag. Sometimes, retrying is not desirable. WDYT?

`etcdadm` add backoff retry flags, like `etcdadm join --retry-steps=5 --retry-duration=2s` ?

chuckha:
> I apologize for my late review! I would prefer making retry/backoff optional using a flag. Sometimes, retrying is not desirable. WDYT?
> 
> lgtm otherwise

Interesting, what use case would you want it to fail because of a transient error? I'd opt to disable the retry with a flag, but curious to hear your reasoning

chuckha:
> > Instead of 1, 1, 2, 2, 4, etc? I think it works fine as is, I'm not sure what you're suggesting here really
> 
> @chuckha Sorrry, i mean join two member at the same time, `etcdadm` backoff retry three times basically, it takes about 4s， so i want to update `DefaultBackOffDuration = 2 * time.Second`, and backoff time is 0, 2, 4, 8, 16s, otherwise if network instablility, join maybe fail.
> 
> > I apologize for my late review! I would prefer making retry/backoff optional using a flag. Sometimes, retrying is not desirable. WDYT?
> 
> `etcdadm` add backoff retry flags, like `etcdadm join --retry-steps=5 --retry-duration=2s` ?

Oh yeah, that's fine with me. My main concern was that etcdadm was failing due to a transient error and needed a retry. If the backoff is reasonable and it ends up joining without failing then I'm happy

pytimer:
@chuckha I update the backoff duration.

dlipovetsky:
> Interesting, what use case would you want it to fail because of a transient error?

Sorry, I did not give enough detail. For example, `MemberAdd` can fail if there is an unstarted member (i.e., a member that was added to the membership list, but that has not started). If you are adding members serially, you might want to treat this as an error, because it indicates some bug with your serialization logic. But if that unstarted member manages to start within the timeout, you might miss the symptom caused by the bug.

> I'd opt to disable the retry with a flag, but curious to hear your reasoning

Agreed; retry can be enabled by default.

pytimer:
> If you are adding members serially, you might want to treat this as an error, because it indicates some bug with your serialization logic. But if that unstarted member manages to start within the timeout, you might miss the symptom caused by the bug.

We record the last error when join fail after retry, is it enough? If unstarted member or occur other error member started within the timeout, `etcdadm` retry and join to cluster after other member running normal, i think this is ok. Because we aren't sure what's happen when join.

During my environment, when i adding members serially, i will check etcd cluster health before add every member, but it maybe occur something when join, such as network not stability, i wants `etcdadm` can auto retry, so retry is useful for me even though add members serially.

I'm not sure above i said if fit the scenario you said.

chuckha:
Do we want to block this PR on adding the flag feature? Totally ok if the answer is yes, I don't have a pressing need for the retry functionality

dlipovetsky:
> Do we want to block this PR on adding the flag feature? Totally ok if the answer is yes, I don't have a pressing need for the retry functionality

Yes, but only because I think adding a flag is straightforward. Retry would be enabled by default, and a user that wants to disable retry would run:
```
etcdadm join --retry=false https://peerURL
```

@pytimer Tell us if you do not have the time to add the flag. We can merge this and add the flag later.

pytimer:
I can add `--retry` flag in this pr.

dlipovetsky:
> I can add `--retry` flag in this pr.

Thank you!

pytimer:
Add `--retry` flag when join etcd member, default value is true.

Issue #56, opened 2018-08-30
Support etcd certificate rotation


sarun87:
The Peer, Client and Server certs we generate today is valid for *one* year. We will need to run `etcdadm reset` followed by `etcdadm join/init` to create new certs within the given year for the etcd cluster to continue functioning.

Cert rollover needs to be implemented.

ambientsound:
My cluster refuses to start today because the etcd certificates expired. I was looking for some tooling to automatically renew them, but it seems I have to do it manually? This cluster was created using kubeadm.

pytimer:
I send a PR to implement certificates rotation, if someone needs, welcome to review it.

PR #147, opened 2019-12-22
Add renew certificates subcommand

pytimer:
The `etcdadm certs renew` support renew certificates with local CA, it will renew the etcd certificates exclude CA.

Now the `etcdadm` create etcd certificates default expiration time is one year. The cluster not work if time is more than one year. I use `kubeadm` create Kubernetes cluster and it also have the same question, however `kubeadm` have renew command to help us make certificates rotation, so i think it is better if `etcdadm`  have this renew command.'

I also found other people have this question, Ref: https://github.com/kubernetes-sigs/etcdadm/issues/56, so i send a PR to add a `renew` subcommand for implement it.

I'm not sure if missing something, if i missing, please tell me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,22 @@ On the machine being removed, run
 etcdadm reset
 ```
 
+### Renewal certificates
+
+1. Ensure the CA certificate and key already exists on the machine.
+
+If not exists, Copy the CA certificate and key from other machine in the cluster.
+
+```
+rsync -avR /etc/etcd/pki/ca.* <Member IP address>:/
+```
+
+2. run command to renew certficates via CA certificate and key.
+
+```
+etcdadm certs renew
+```
+
 ## Advanced Usage
 
 ### Creating a new cluster from a snapshot
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -46,7 +46,7 @@ func GetDefaultCertList() []string {
 }
 
 // RenewUsingLocalCA replaces certificates with new certificates signed by the CA.
-func RenewUsingLocalCA(pkiDir, name string) (bool, error) {
+func RenewUsingLocalCA(pkiDir, name string, csrOnly bool) (bool, error) {
 	// reads the current certificate
 	cert, err := pkiutil.TryLoadCertFromDiskIgnoreExpirationDate(pkiDir, name)
 	if err != nil {
@@ -55,6 +55,17 @@ func RenewUsingLocalCA(pkiDir, name string) (bool, error) {
 
 	// extract the certificate config
 	cfg := certToConfig(cert)
+
+	if csrOnly {
+		csr, key, err := pkiutil.NewCSRAndKey(&cfg)
+		if err != nil {
+			return false, errors.Wrapf(err, "failed to renew csr %s", name)
+		}
+		if err := writeCSRFilesIfNotExist(pkiDir, name, csr, key); err != nil {
+			return false, errors.Wrapf(err, "failed to write new csr %s", name)
+		}
+		return true, nil
+	}
 
 	// reads the CA
 	caCert, caKey, err := loadCertificateAuthority(pkiDir, constants.EtcdCACertAndKeyBaseName)
@@ -426,6 +437,30 @@ func writeKeyFilesIfNotExist(pkiDir string, baseName string, key *rsa.PrivateKey
 			return fmt.Errorf("failure while saving %s public key: %v", baseName, err)
 		}
 		fmt.Printf("[certificates] Generated %s key and public key.\n", baseName)
+	}
+
+	return nil
+}
+
+func writeCSRFilesIfNotExist(csrDir string, baseName string, csr *x509.CertificateRequest, key *rsa.PrivateKey) error {
+	if pkiutil.CSROrKeyExist(csrDir, baseName) {
+		_, _, err := pkiutil.TryLoadCSRAndKeyFromDisk(csrDir, baseName)
+		if err != nil {
+			return errors.Wrapf(err, "%s CSR existed but it could not be loaded properly", baseName)
+		}
+
+		fmt.Printf("[certs] Using the existing %q CSR\n", baseName)
+	} else {
+		// Write .key and .csr files to disk
+		fmt.Printf("[certs] Generating %q key and CSR\n", baseName)
+
+		if err := pkiutil.WriteKey(csrDir, baseName, key); err != nil {
+			return errors.Wrapf(err, "failure while saving %s key", baseName)
+		}
+
+		if err := pkiutil.WriteCSR(csrDir, baseName, csr); err != nil {
+			return errors.Wrapf(err, "failure while saving %s CSR", baseName)
+		}
 	}
 
 	return nil
--- a/certs/certs.go
+++ b/certs/certs.go
@@ -26,6 +26,7 @@ import (
 	"crypto/x509"
 	"fmt"
 
+	"github.com/pkg/errors"
 	log "sigs.k8s.io/etcdadm/pkg/logrus"
 
 	certutil "k8s.io/client-go/util/cert"
@@ -33,6 +34,59 @@ import (
 	"sigs.k8s.io/etcdadm/certs/pkiutil"
 	"sigs.k8s.io/etcdadm/constants"
 )
+
+// GetDefaultCertList returns all of the certificates etcdadm requires to function.
+func GetDefaultCertList() []string {
+	return []string{
+		constants.EtcdServerCertAndKeyBaseName,
+		constants.EtcdPeerCertAndKeyBaseName,
+		constants.EtcdctlClientCertAndKeyBaseName,
+		constants.APIServerEtcdClientCertAndKeyBaseName,
+	}
+}
+
+// RenewUsingLocalCA replaces certificates with new certificates signed by the CA.
+func RenewUsingLocalCA(pkiDir, name string) (bool, error) {
+	// reads the current certificate
+	cert, err := pkiutil.TryLoadCertFromDiskIgnoreExpirationDate(pkiDir, name)
+	if err != nil {
+		return false, err
+	}
+
+	// extract the certificate config
+	cfg := certToConfig(cert)
+
+	// reads the CA
+	caCert, caKey, err := loadCertificateAuthority(pkiDir, constants.EtcdCACertAndKeyBaseName)
+	if err != nil {
+		return false, err
+	}
+
+	// create a new certificate with the same config
+	newCert, newKey, err := pkiutil.NewCertAndKey(caCert, caKey, cfg)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to renew certificate %s", name)
+	}
+
+	// writes the new certificate to disk
+	if err := pkiutil.WriteCertAndKey(pkiDir, name, newCert, newKey); err != nil {
+		return false, errors.Wrapf(err, "failed to write new certificate %s", name)
+	}
+
+	return true, nil
+}
+
+func certToConfig(cert *x509.Certificate) certutil.Config {
+	return certutil.Config{
+		CommonName:   cert.Subject.CommonName,
+		Organization: cert.Subject.Organization,
+		AltNames: certutil.AltNames{
+			IPs:      cert.IPAddresses,
+			DNSNames: cert.DNSNames,
+		},
+		Usages: cert.ExtKeyUsage,
+	}
+}
 
 // CreatePKIAssets will create and write to disk all PKI assets necessary to establish the control plane.
 // If the PKI assets already exists in the target folder, they are used only if evaluated equal; otherwise an error is returned.
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/etcdadm/certs"
+	"sigs.k8s.io/etcdadm/constants"
+	log "sigs.k8s.io/etcdadm/pkg/logrus"
+)
+
+// newCmdCerts returns main command for certs phase
+func newCmdCerts() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "certs",
+		Short: "Commands related to handling etcdadm certificates",
+	}
+
+	cmd.AddCommand(newCmdCertsRenewal())
+
+	return cmd
+}
+
+func newCmdCertsRenewal() *cobra.Command {
+	var certificatesDir string
+	cmd := &cobra.Command{
+		Use:   "renew",
+		Short: "Renew certificates for a etcd cluster",
+		Run: func(cmd *cobra.Command, args []string) {
+			for _, name := range certs.GetDefaultCertList() {
+				log.Printf("[renew] Renew etcd %s certificate.", name)
+				renewed, err := certs.RenewUsingLocalCA(certificatesDir, name)
+				if err != nil {
+					log.Fatalf("[renew] Error renew certificate %s: %v", name, err)
+				}
+				if !renewed {
+					log.Fatalf("Certificates %s can't be renewed.\n", name)
+				}
+			}
+			log.Println("The etcd certificates have been renewed successfully!")
+			log.Warnln("If kube-apiserver which version less than v1.10.0 is running, restart it so that it uses the renewed etcd client certificate.")
+		},
+	}
+	cmd.PersistentFlags().StringVar(&certificatesDir, "certs-dir", constants.DefaultCertificateDir, "certificates directory")
+
+	return cmd
+}
+
+func init() {
+	rootCmd.AddCommand(newCmdCerts())
+}
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -37,13 +37,14 @@ func newCmdCerts() *cobra.Command {
 
 func newCmdCertsRenewal() *cobra.Command {
 	var certificatesDir string
+	var csrOnly bool
 	cmd := &cobra.Command{
 		Use:   "renew",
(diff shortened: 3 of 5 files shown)


pytimer:
This command should be run on the certificate expirated node with CA. Before run this command, you should copy CA to the node, and then run `renew` command, it will renew the etcd certificates on the node.

chuckha:
> This command should be run on the certificate expirated node with CA. Before run this command, you should copy CA to the node, and then run `renew` command, it will renew the etcd certificates on the node.

My questions were mostly to help guide some documentation fore this feature. Perhaps a new section in the README? Or maybe even in the command help?

pytimer:
@chuckha I update the README.md in this pr, you can review it. Thanks.

README.md
```diff
@@ -60,6 +60,22 @@ On the machine being removed, run
 etcdadm reset
 ```
 
+### Renewal certificates
```
chuckha: This took me a second to understand. We aren't renewing the CA, we're renewing the certs generated with the CA. 

this should be a little more friendly, how about something like this:


> Certificates are set to expire after <1 year?> by default. 
> To renew certificates, first ensure every node in the cluster has the CA root certificate. Then, on each node in, run `etcdadm certs renew`.

dlipovetsky: +1 to @chuckha's suggestion.

I think it might help to explicitly mention that there are two certificates: peer and server.

certs/certs.go
```diff
@@ -34,6 +35,60 @@ import (
 	"sigs.k8s.io/etcdadm/constants"
 )
 
+// GetDefaultCertList returns all of the certificates etcdadm requires to function.
+func GetDefaultCertList() []string {
+	return []string{
+		//constants.EtcdCACertAndKeyBaseName,
```
dlipovetsky: Remove commented code?

certs/certs.go
```diff
@@ -34,6 +35,60 @@ import (
...
+		constants.EtcdServerCertAndKeyBaseName,
+		constants.EtcdPeerCertAndKeyBaseName,
+		constants.EtcdctlClientCertAndKeyBaseName,
+		constants.APIServerEtcdClientCertAndKeyBaseName,
+	}
+}
+
+// RenewUsingLocalCA executes certificate renewal using local certificate authorities for generating new certs.
```
dlipovetsky:
```suggestion
// RenewUsingLocalCA replaces certificates with new certificates signed by the CA.
```

cmd/certs.go
```diff
@@ -0,0 +1,46 @@
+package cmd
```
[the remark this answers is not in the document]
pytimer: I will add license header late.

certs/certs.go
```diff
@@ -34,6 +35,60 @@ import (
 	"sigs.k8s.io/etcdadm/constants"
 )
 
+// GetDefaultCertList returns all of the certificates etcdadm requires to function.
+func GetDefaultCertList() []string {
+	return []string{
+		//constants.EtcdCACertAndKeyBaseName,
+		constants.EtcdServerCertAndKeyBaseName,
```
dlipovetsky: How does replacing a certificate affect the client that uses the certificates?

1. etcd peer and server certificates: etcd will check and, if needed, [reload the certificate on every connection](https://github.com/etcd-io/etcd/pull/7829).
2. etcdctl client certificate: this is not a long-running process, and presumably etcdctl reads certificates when it starts. If it fails, the user can always rerun it.
3. **kube-apiserver: I'm not sure if it will reload its etcd client certificate. Does it need to be restarted? **

README.md
pytimer: I will update the document.

certs/certs.go
pytimer: 1. Etcd should restart as far as I know.

2. etcdctl every time use the new certificate s , so I think renew not effect it.

3. If the kube-apisever client crt changed, I think kube-apisever maybe need to restart. I'm not tests this case, so i'm not sure. I will tests it.

certs/certs.go
pytimer: OK, I will remove it.

certs/certs.go
```diff
@@ -34,6 +35,60 @@ import (
...
+// RenewUsingLocalCA executes certificate renewal using local certificate authorities for generating new certs.
+func RenewUsingLocalCA(pkiDir, name string) (bool, error) {
+	// reads the current certificate
+	cert, err := pkiutil.TryLoadCertFromDiskIgnoreExpirationDate(pkiDir, name)
+	if err != nil {
+		return false, err
+	}
+
+	// extract the certificate config
+	cfg := certToConfig(cert)
```
dlipovetsky: For `etcdadm certs renew` this means that the source of truth is the certificate on disk.. However, for `etcdadm init` and `etcdadm join`, the source of truth is [etcdadm](https://github.com/kubernetes-sigs/etcdadm/blob/bec4908b5008104cf1637644335a6b0c986bcf5a/certs/certs.go#L190).

I think that the source of truth should be consistent across all commands.

In fact, if  `etcdadm certs renew` uses etcdadm as the source of truth, then it can just create new certificates using the existing functions (`CreateEtcdServerCertAndKeyFiles`, `CreateEtcdPeerCertAndKeyFiles`,  `CreateEtcdctlClientCertAndKeyFiles`, and `CreateAPIServerEtcdClientCertAndKeyFiles`).

WDYT?

certs/certs.go
dlipovetsky: Thank you!

If kube-apiserver does need to be restarted, let's document that. Perhaps even write a message to the user.

certs/certs.go
pytimer: You mean that maybe the disk certificates not the source of truth? 

I understand your said, But if we use createXXX to recreate the new certificates, is it include CA certificate?

dlipovetsky: > You mean that maybe the disk certificates not the source of truth?

That's right; I believe that the certificates on disk should not be the source of truth.

When etcdadm creates certificates the first time, it determines their properties. When it renews them, it should set those properties the same way. It should not pay attention to the properties in the certificates on disk.

> I understand your said, But if we use createXXX to recreate the new certificates, is it include CA certificate?

I believe each of the functions I mentioned reads the CA from disk.

README.md
dlipovetsky: Thank you @pytimer!

certs/certs.go
pytimer: Do not believe the certificates on disk is right, but if we recreate certificates, this command name still call `renew`, or change to `etcdadm certs recreate` ?

dlipovetsky: Great question! I think `renew` is correct, based on the RFC:

> This subcomponent is used to describe the following elements related to certificate renewal. Certificate renewal means the issuance of a new certificate to the subscriber without changing the subscriber or other participant's public key or any other information in the certificate:
-- https://tools.ietf.org/html/rfc3647 *

However, if I'm reading the definition correctly, etcdadm must generate a new certificate using its current key on disk. That means etcdadm needs to read the key from disk, but not the certificate.

(I continue to believe that etcdadm should set the certificate properties, rather than read them from the certificate on disk.)

\* I found this via https://security.stackexchange.com/questions/150907/x-509-certificate-renew-vs-rekey

pytimer: I think this renew implementation more like rekey, but it also use certificate  properties. If my understanding correct, you mean that etcdadm not use old certificate properties on the disk?

dlipovetsky: I agree; this PR right now implements "rekey" according to the RFC.

I don't know whether we want "renew" or "rekey;" I need to learn the trade-offs and use cases.

I decided to check what kubeadm does. Looks like kubeadm does in fact read the properties of the existing certificates:
> Note: alpha certs renew uses the existing certificates as the authoritative source for attributes (Common Name, Organization, SAN, etc.) instead of the kubeadm-config ConfigMap. It is strongly recommended to keep them both in sync.
-- https://v1-17.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#renew-certificates-with-the-kubernetes-certificates-api

I wonder why that is.

pytimer: I see kubeadm read certificate properties, i think maybe etcdadm init or join can add custom SANs, if we don't read properties, maybe we missing some SANs.

dlipovetsky: >  i think maybe etcdadm/kubeadm init or join can add custom SANs, if we don't read properties, maybe we missing some SANs.

That's a great point.

dlipovetsky:
Thanks for helping me understand your approach @pytimer . I'm ok with etcdadm reading the properties from the certs on disk.

The only question left is https://github.com/kubernetes-sigs/etcdadm/pull/147#discussion_r367170736.  Otherwise this looks good to merge.

pytimer:
> The only question left is #147 (comment). Otherwise this looks good to merge.

Sorry, because of have some things, i have no locally Kubernetes cluster before, i will testing and reply it coming soon.

pytimer:
@dlipovetsky I testing kube-apiserver case about https://github.com/kubernetes-sigs/etcdadm/pull/147#discussion_r367151574 , kube-apiserver needs to restart and load new etcd client certificates.

I will update document and code, warning users to pay attention.

cmd/certs.go
```diff
@@ -0,0 +1,65 @@
...
+				log.Printf("[renew] Renew etcd %s certificate.", name)
+				renewed, err := certs.RenewUsingLocalCA(certificatesDir, name)
+				if err != nil {
+					log.Fatalf("[renew] Error renew certificate %s: %v", name, err)
+				}
+				if !renewed {
+					log.Fatalf("Certificates %s can't be renewed.\n", name)
+				}
+			}
+			log.Println("Your etcd certificates has renew successfully!")
```
dlipovetsky:
```suggestion
			log.Println("The etcd certificates have been renewed successfully!")
```

cmd/certs.go
```diff
@@ -0,0 +1,65 @@
...
+			log.Warnln("After renew etcd certificates, you need to restart client(kube-apiserver) which use apiserver-etcd-client certificate.")
```
dlipovetsky:
```suggestion
			log.Warnlng("If kube-apiserver is running, restart it so that it uses the renewed etcd client certificate.")
```

dlipovetsky:
> Worked for me, but I'm surprised at the lack of cluster level health check/info through etcdadm.

@chuckha What did you want etcdadm to do?

Also, what do you think of the PR in it stands?

dlipovetsky:
> @dlipovetsky I testing kube-apiserver case about [#147 (comment)](https://github.com/kubernetes-sigs/etcdadm/pull/147#discussion_r367151574) , kube-apiserver needs to restart and load new etcd client certificates.
> 
> I update document and code, warning users to pay attention. The code is: https://github.com/kubernetes-sigs/etcdadm/pull/147/files#diff-1094222d3259d85ad53cc70f8bcc0816R54

I know I said earlier that kube-apisever might need to be restarted to reload its etcd client cert. But is that what you see in the logs? The reason I ask is: it looks like kube-apiserver should reload the certificates automatically. That's because kube-apiserver uses the [etcd client's TLS config](https://github.com/kubernetes/kubernetes/blob/9ead9373f350c7ad438257a9e8b8977a67b900eb/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L104), which implements [certificate reloading](https://github.com/etcd-io/etcd/pull/7829).

pytimer:
> I know I said earlier that kube-apisever might need to be restarted to reload its etcd client cert. But is that what you see in the logs? The reason I ask is: it looks like kube-apiserver should reload the certificates automatically. That's because kube-apiserver uses the [etcd client's TLS config](https://github.com/kubernetes/kubernetes/blob/9ead9373f350c7ad438257a9e8b8977a67b900eb/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L104), which implements [certificate reloading](https://github.com/etcd-io/etcd/pull/7829).

I see kube-apiserver log print error, `remote error: tls: bad certificate`. I'm not sure kube-apiserver occur this error because of renew certificates, i will continue tests. If someone know, please point out.

dlipovetsky:
> I see kube-apiserver log print error, `remote error: tls: bad certificate`. I'm not sure kube-apiserver occur this error because of renew certificates, i will continue tests. If someone know, please point out.

I'll look this week, too. It's not a blocking question, but it's important to know. Depending on what we find, we may want to open a kube-apiserver issue.

pytimer:
@dlipovetsky I test the https://github.com/kubernetes-sigs/etcdadm/pull/147#discussion_r367151574 user case. Renew the `apiserver-etcd-client.crt` and `apiserver-etcd-client.key`, we just need to replace the old certificates used by kube-apiserver, no need to restart kube-apiserver.

>  That's because kube-apiserver uses the [etcd client's TLS config](https://github.com/kubernetes/kubernetes/blob/9ead9373f350c7ad438257a9e8b8977a67b900eb/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L104), which implements [certificate reloading](https://github.com/etcd-io/etcd/pull/7829).

I look etcd and kubernetes code about it, etcd support reload TLS certificates after 3.2.0, and kubernetes use etcd v3.2.x version after v1.10.0, so should we warning user restart kube-apiserver when use kubernetes version before v1.10.0?

kube-apiserver logs:

```diff
+W0508 01:30:44.652566       1 clientconn.go:1120] grpc: addrConn.createTransport failed to connect to {https://10.33.41.239:2379 0  <nil>}. Err :connection error: desc = "transport: authentication handshake failed: remote error: tls: bad certificate". Reconnecting...
+W0508 01:30:44.714464       1 clientconn.go:1120] grpc: addrConn.createTransport failed to connect to {https://10.33.41.239:2379 0  <nil>}. Err :connection error: desc = "transport: authentication handshake failed: remote error: tls: bad certificate". Reconnecting...
# After use new certificates replace old certificates.
I0508 01:30:48.312943       1 trace.go:116] Trace[12652262]: "GuaranteedUpdate etcd3" type:*coordination.Lease (started: 2021-05-08 01:30:38.317134174 +0000 UTC m=+60004.260302142) (total time: 9.995741601s):
Trace[12652262]: [9.995741601s] [9.99522434s] END
E0508 01:30:48.313020       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"context canceled"}
E0508 01:30:48.313190       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"context canceled"}
E0508 01:30:48.313317       1 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s:"context canceled"}
I0508 01:30:48.313559       1 trace.go:116] Trace[77263109]: "Update" url:/apis/coordination.k8s.io/v1/namespaces/kube-node-lease/leases/k8s-234,user-agent:kubelet/v1.17.2 (linux/amd64) kubernetes/59603c6,client:10.33.41.239 (started: 2021-05-08 01:30:38.316831689 +0000 UTC m=+60004.259999604) (total time: 9.996673434s):
Trace[77263109]: [9.996673434s] [9.996458176s] END
I0508 01:30:48.314439       1 trace.go:116] Trace[1173260492]: "Create" url:/api/v1/namespaces/kube-system/events,user-agent:kubelet/v1.17.2 (linux/amd64) kubernetes/59603c6,client:10.33.41.239 (started: 2021-05-08 01:30:36.957526238 +0000 UTC m=+60002.900694114) (total time: 11.356876436s):
```

Avatat:
@pytimer, thank you for your effort and for adding this feature!

Can you add `--csr-only` option [known from kubeadm](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#renew-certificates-with-external-ca)? It's a useful addition for those who use external CA :)

pytimer:
@Avatat I'm not understand kubeadm how to renew certificates via  `csr-only`, i will check what kubeadm does.

I think etcdadm support renew certificates with  local CA first, then implement external CA, i want to do it in the new PR. 

What do you think? @Avatat @dlipovetsky

Avatat:
When `--csr-only` is used, etcdadm should generate only keys to the certificates and [CSR's](https://en.wikipedia.org/wiki/Certificate_signing_request).

Here is how it's done [in kubeadm](https://github.com/kubernetes/kubernetes/blob/4e79344501bfc5d3bc48caa12c41b9246d90850d/cmd/kubeadm/app/cmd/phases/init/certs.go#L281).
I see, that etcdadm and kubeadm share the same `pkiutil` functions.

pytimer:
> Here is how it's done in kubeadm.

Thanks, i will do it next week.

Avatat:
Any updates? :)

pytimer:
Sorry, i am working on the kubernetes performance testing at latest two weeks. I plan to do this feature at next week.

pytimer:
@Avatat I add `--csr-only` flags to create csr instead of generate certificates when using external CA, you can tests it with the latest commit code. If anything wrong, welcome to point out.

Avatat:
I'm sorry. I will check this feature next week.

ntaylor1781:
Can we reopen this? This is a needed feature. I think https://github.com/kubernetes-sigs/etcdadm/issues/179 is a great long term goal, but it will be able to utilize this work

ntaylor1781:
A comment on kube-apiserver. It will pick up the changes to the ETCD client cert, and automatically use it on new connections.

pytimer:
> A comment on kube-apiserver. It will pick up the changes to the ETCD client cert, and automatically use it on new connections.

I use k8s v1.10.0, found kube-apiserver can not automatically use the new client cert, need to restart it.

ntaylor1781:
Good point. The functionality was added in the etcd client version 3.2.0, though there was a bug where it did not work properly when addressed with an IP until 3.2.19.  K8S version 1.11 and 1.10 should work if addressed with a domain. Version 1.12 and up will always reload.

ntaylor1781:
Can we re-open this and get it reviewed?

Avatat:
Finally, I tried it and the renew feature works good!

The only problem I have is CSR mode:
```
# etcdadm certs renew --csr-only
INFO[0000] [renew] Renew etcd server certificate.       
FATA[0000] [renew] Error renew certificate server: failed to write new csr server: server CSR existed but it could not be loaded properly: couldn't load the certificate request /etc/etcd/pki/server.csr: failed to read file: open /etc/etcd/pki/server.csr: no such file or directory
```
But there is no `server.csr` file:
```
# ls
apiserver-etcd-client.crt  apiserver-etcd-client.key  ca.crt  ca.key  etcdctl-etcd-client.crt  etcdctl-etcd-client.key	peer.crt  peer.key  server.crt	server.key
```


If you want to use this feature, clone @pytimer repo, build from `cert` branch, and use built binary:
```
git clone -b cert https://github.com/pytimer/etcdadm.git
cd etcdadm
make container-build
./etcdadm certs
```

sfgroups-k8s:
Hi,
Any plan to include this cert renewal code  https://github.com/pytimer/etcdadm/tree/cert to this etcdadm code?

Thanks

Issue #108, opened 2019-06-24
create a etcdadm versioned configuration format


neolit123:
we discussed during the meeting of 24th of June 2019 that etcdadm needs a configuration format, that follows the Kuberentes API machinery.

https://docs.google.com/document/d/1b_J0oBvi9lL0gsPgTOrCw1Zlx3e7BYEuXnB3d2S15pA/edit#

this will allows us to mange differences between etcd version WRT to etcdctl flags and etcd API changes between releases.

the kubeadm etcd types can be used as a starting point.
https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2#ExternalEtcd
https://godoc.org/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2#LocalEtcd

if anybody has ideas please add them in this ticket.

/priority important-soon
/kind feature

/assign @justinsb 
cc @mattymo

xrmzju:
hi, i've implement a internal version etcdadm versioned config and works fine in our situation, i wrote a brief design doc for the implemention @dlipovetsky @justinsb would you mind take a look at this?(
https://docs.google.com/document/d/1zpztxZ3t7AjoXUrIefzbnqyVLATwATqrvFaRaR-8Xm8)

dlipovetsky:
Thanks, @xrmzju . I'll take a look soon.

PR #154, opened 2020-04-05
Expose the peer and server certificate SAN list as command line flags

mdaniel:
Otherwise when using `etcdadm init` behind a load balancer (or DNS CNAME), it is impossible for the generated certs to work for the newly initialized member

I am aware of the work in #108 which will almost certainly supersede these flags, but "perfect enemy good etc etc" and since the missing flags were very short, I thought I'd send the PR along for consideration

Goland was whining about those ignored `error` results, so I "left the campsite cleaner" on those, but they're in a separate commit if you don't want those changes


cmd/init.go
```diff
@@ -153,14 +153,17 @@ var initCmd = &cobra.Command{
...
-	initCmd.PersistentFlags().DurationVar(&etcdAdmConfig.DownloadConnectTimeout, "download-connect-timeout", constants.DefaultDownloadConnectTimeout, "Maximum time in seconds that you allow the connection to the server to take.")
-	initCmd.PersistentFlags().StringArrayVar(&etcdAdmConfig.EtcdDiskPriorities, "disk-priorities", constants.DefaultEtcdDiskPriorities, "Setting etcd disk priority")
+	flags := initCmd.PersistentFlags()
+	flags.StringVar(&etcdAdmConfig.Name, "name", "", "etcd member name")
+	flags.StringVar(&etcdAdmConfig.Version, "version", constants.DefaultVersion, "etcd version")
+	flags.StringVar(&etcdAdmConfig.ReleaseURL, "release-url", constants.DefaultReleaseURL, "URL used to download etcd")
+	flags.StringVar(&etcdAdmConfig.CertificatesDir, "certs-dir", constants.DefaultCertificateDir, "certificates directory")
+	flags.StringSliceVar(&etcdAdmConfig.ServerCertSANs, "server-cert-extra-sans", etcdAdmConfig.ServerCertSANs, "optional extra Subject Alternative Names for the etcd server signing cert, can be multiple comma separated DNS names or IPs")
```
dlipovetsky, line 163: Does this flag already do what you need for server certs?

cmd/init.go
```diff
@@ -153,14 +153,17 @@ var initCmd = &cobra.Command{
...
+	flags.StringVar(&etcdAdmConfig.InstallDir, "install-dir", constants.DefaultInstallDir, "install directory")
+	flags.StringVar(&etcdAdmConfig.Snapshot, "snapshot", "", "Etcd v3 snapshot file used to initialize member")
+	flags.BoolVar(&etcdAdmConfig.SkipHashCheck, "skip-hash-check", false, "Ignore snapshot integrity hash value (required if copied from data directory)")
+	flags.DurationVar(&etcdAdmConfig.DownloadConnectTimeout, "download-connect-timeout", constants.DefaultDownloadConnectTimeout, "Maximum time in seconds that you allow the connection to the server to take.")
+	flags.StringArrayVar(&etcdAdmConfig.EtcdDiskPriorities, "disk-priorities", constants.DefaultEtcdDiskPriorities, "Setting etcd disk priority")
+	flags.StringArrayVar(&etcdAdmConfig.PeerCertSANs, "peer-cert-sans", etcdAdmConfig.PeerCertSANs, "Setting etcd peer certificate SAN list")
```
dlipovetsky, line 169: Can you please explain the need for extra peer SANs?

dlipovetsky:
Thank you for fixing the unchecked errors.

cmd/init.go
mdaniel: Damn, I'm sorry -- I got bitten by "too much pasting, not enough reading" since I had that patch against the latest released version of etcdadm (v0.1.2) and I didn't _read_ before pasting those lines

Given that both `init.go` and `join.go` now correctly contain `ServerCertSANs`, plus the perfectly valid "why expose PeerSANs" question below, then I think this PR can be closed.

---

That said, it looks like the ServerCertSANs just missed the v0.1.2 release window, so is a fresh release of `etcdadm` coming? And if so, could I beg and plead to have that next release to have the usual go binaries attached to it, in order to keep one from having to build `etcdadm` locally?

cmd/init.go
mdaniel: It wasn't "need," as much as the Config had the field and it seemed appropriate to surface it alongside its Server SANs friend; I also freely admit that I don't have an iron grip on which processes use which certificates for which purpose in order to say with high confidence that they would never try to use peer.crt to contact their peers through the LB

dlipovetsky: Got it. Once there's a concrete use case for extra SANs for the peer certificate, I think it would be great to add that as an option (and flag). But until then, I would humbly request not providing the option.

cmd/init.go
dlipovetsky: Thanks for explaining. I will make an effort this week to include go binaries in the release.

Issue #157, opened 2020-04-26
Adding support for --listen-metrics-url parameter 


suhrud-kumar:
etcd has support for providing separate address for metrics endpoint.

This can be useful if the standard endpoint is configured with mutual (client) TLS authentication, but a load balancer or monitoring service still needs access to the metrics or health check. (etcd api will be served over https but /metrics endpoint will served over http )

Since etcdadm creates etcd in TLS be default, this will be useful for folks who don't to add etcd certs in their monitoring systems (for security reasons).

Currently we implemented a small hack to get around this, but it would be great if this is supported by etcdadm.

Let me know what you guys think, I can try to implement this with some guidance.

References:
https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/configuration.md#--listen-metrics-urls
https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/monitoring.md#metrics-endpoint

dlipovetsky:
I agree, exposing the metrics endpoint would be useful. It's a great idea, thanks for suggesting it @suhrud-kumar!

> Since etcdadm creates etcd in TLS be default, this will be useful for folks who don't to add etcd certs in their monitoring systems (for security reasons).

I'm skeptical of this motivation. Can you explain why someone might not want to "add etcd certs in their monitoring system?" (Thinking out loud: The etcd metrics endpoint could be used by a "pull" monitoring system (like Prometheus), and that system would only need a client certificate. The metrics client certificate, as far as I can tell, would have to be the same as the normal client certificate. So the monitoring system would have access to etcd data. I see why that's not desirable.)

I found some related discussion here: https://github.com/etcd-io/etcd/issues/8060#issuecomment-419462582  


If you're able to work on it, please use the "/lifecycle active" command. If you have any implementation questions, please ask in #etcdadm channel in Kubernetes slack.

/priority important-longterm

suhrud-kumar:
> I agree, exposing the metrics endpoint would be useful. It's a great idea, thanks for suggesting it @suhrud-kumar!

Just to clarify, metrics endpoint will be exposed by default but in https (if installed using etcdadm), by adding this parameter an additional metrics endpoint will be exposed in http

> > Since etcdadm creates etcd in TLS be default, this will be useful for folks who don't to add etcd certs in their monitoring systems (for security reasons).
> 
[...]

PR #160, opened 2020-05-08
added support for --listen-metrics-url parameter

suhrud-kumar:
support for https://github.com/kubernetes-sigs/etcdadm/issues/157

@dlipovetsky 
Added new flag --listen-metrics-url

- If user passes a URL to this parameter, we'll use that URL to expose additional metrics endpoint in http
- By default (without passing any extra params) this will expose /metrics endpoint on http://client-ip:9379

I can't seem to make this optional without messing up with existing code structure. Let me know if I need to make any changes

### DIFF
--- a/cmd/join.go
+++ b/cmd/join.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/etcdadm/etcd"
 	"sigs.k8s.io/etcdadm/initsystem"
 	"sigs.k8s.io/etcdadm/service"
+	"sigs.k8s.io/etcdadm/util"
 
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -235,4 +236,5 @@ func init() {
 	joinCmd.PersistentFlags().StringVar(&etcdAdmConfig.InstallDir, "install-dir", constants.DefaultInstallDir, "install directory")
 	joinCmd.PersistentFlags().StringArrayVar(&etcdAdmConfig.EtcdDiskPriorities, "disk-priorities", constants.DefaultEtcdDiskPriorities, "Setting etcd disk priority")
 	joinCmd.PersistentFlags().BoolVar(&etcdAdmConfig.Retry, "retry", true, "Enable or disable backoff retry when join etcd member to cluster")
+	joinCmd.PersistentFlags().Var(util.URLValue{URL: &etcdAdmConfig.AdditionalMetricsURL}, "listen-metrics-url", "URL to expose additional /metrics endpoint without TLS")
 }
--- a/util/url_value.go
+++ b/util/url_value.go
@@ -1,0 +1,49 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"net/url"
+)
+
+// URLValue struct definition
+type URLValue struct {
+	URL *url.URL
+}
+
+// String reassembles the URL into a valid URL string
+func (s URLValue) String() string {
+	if s.URL != nil {
+		return s.URL.String()
+	}
+	return ""
+}
+
+// Set sets the argument passed to URL Value
+func (s URLValue) Set(val string) error {
+	if u, err := url.Parse(val); err != nil {
+		return err
+	} else {
+		*s.URL = *u
+	}
+	return nil
+}
+
+// Type returns type of struct
+func (s URLValue) Type() string {
+	return "URLValue"
+}
--- a/util/url_value.go
+++ b/util/url_value.go
@@ -35,11 +35,14 @@ func (s URLValue) String() string {
 
 // Set sets the argument passed to URL Value
 func (s URLValue) Set(val string) error {
-	if u, err := url.Parse(val); err != nil {
+
+	var u *url.URL
+	var err error
+
+	if u, err = url.ParseRequestURI(val); err != nil {
 		return err
-	} else {
-		*s.URL = *u
 	}
+	*s.URL = *u
 	return nil
 }
 
--- a/apis/config.go
+++ b/apis/config.go
@@ -68,7 +68,8 @@ type EtcdAdmConfig struct {
 	AdvertiseClientURLs      URLList
 	ListenClientURLs         URLList
 
-	LoopbackClientURL url.URL
+	LoopbackClientURL    url.URL
+	AdditionalMetricsURL url.URL
 
 	// ServerCertSANs sets extra Subject Alternative Names for the etcd server signing cert.
 	ServerCertSANs []string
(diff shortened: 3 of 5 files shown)


dlipovetsky:
> I can't seem to make this optional without messing up with existing code structure. Let me know if I need to make any changes

I want to make sure I understand what you mean here. When you say"make this optional," are you saying "make the use of a separate metrics endpoint optional?" By optional, I mean: If the `--listen-metrics-url` flag is not used, then etcd does not use a separate metrics endpoint.

util/url_value.go
```diff
@@ -0,0 +1,49 @@
...
+		return s.URL.String()
+	}
+	return ""
+}
+
+// Set sets the argument passed to URL Value
+func (s URLValue) Set(val string) error {
+	if u, err := url.Parse(val); err != nil {
+		return err
+	} else {
```
dlipovetsky, line 40: The CI linter failed: `util/url_value.go:40:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)`

cmd/join.go
```diff
@@ -235,4 +236,5 @@ func init() {
 	joinCmd.PersistentFlags().StringVar(&etcdAdmConfig.InstallDir, "install-dir", constants.DefaultInstallDir, "install directory")
 	joinCmd.PersistentFlags().StringArrayVar(&etcdAdmConfig.EtcdDiskPriorities, "disk-priorities", constants.DefaultEtcdDiskPriorities, "Setting etcd disk priority")
 	joinCmd.PersistentFlags().BoolVar(&etcdAdmConfig.Retry, "retry", true, "Enable or disable backoff retry when join etcd member to cluster")
+	joinCmd.PersistentFlags().Var(util.URLValue{URL: &etcdAdmConfig.AdditionalMetricsURL}, "listen-metrics-url", "URL to expose additional /metrics endpoint without TLS")
```
dlipovetsky, line 239: What happens if the user passes in a string that isn't a valid URL?

util/url_value.go
```diff
@@ -0,0 +1,49 @@
+/*
```
dlipovetsky, line 1: Are the `Set` and `Type` receivers used anywhere?

util/url_value.go
suhrud-kumar: updated as per suggestion

cmd/join.go
suhrud-kumar: we use url.ParseRequestURI() function in URLValue Set function to parse it into a valid URL, if not it returns an error.

[root@ip-172-31-17-207 tmp]# ./etcdadm init --listen-metrics-url http//1.1.1.1
Error: invalid argument "http//1.1.1.1" for "--listen-metrics-url" flag: parse http//1.1.1.1: invalid URI for request

[root@ip-172-31-17-207 tmp]# ./etcdadm init --listen-metrics-url 1.1.1.1
Error: invalid argument "1.1.1.1" for "--listen-metrics-url" flag: parse 1.1.1.1: invalid URI for request

util/url_value.go
suhrud-kumar: Set is used to set argument passed to the variable.

Since URLValue is an implementation of Value interface in pflag, all three functions are required - https://godoc.org/github.com/spf13/pflag#Value

suhrud-kumar:
> By optional, I mean: If the `--listen-metrics-url` flag is not used, then etcd does not use a separate metrics endpoint.

Yes, thats what I meant.

Just to clarify, with current implementation we have below two options

etcdadm init  -> additional http /metrics endpoint on http://\<etcd-client-ip>:defaultport
etcdadm init --listen-metrics-url http://ip:port ->  additional http /metrics endpoint on user requested URL

but we don't have an option to disable this.

Initially I wanted to design something like below

etcdadm init  -> no additional http /metrics endpoint
etcdadm init --listen-metrics-url -> additional http /metrics endpoint on http://\<client-ip>:defaultport
etcdadm init --listen-metrics-url  http://ip:port ->  additional http /metrics endpoint on use requested URL

Since etcdadm is for making etcd operations simple, I think we can accommodate this design consideration.

suhrud-kumar:
I signed it

util/url_value.go
dlipovetsky: Thanks. I missed this.

cmd/join.go
dlipovetsky: Thanks for explaining. I overlooked the use of `.Var`.

The UX looks good! (Do you think it would be nice if the error gives an example of a valid URI?)

dlipovetsky:
> etcdadm init -> no additional http /metrics endpoint
> etcdadm init --listen-metrics-url -> additional http /metrics endpoint on http://<client-ip>:defaultport
> etcdadm init --listen-metrics-url http://ip:port -> additional http /metrics endpoint on use requested URL

I favor this, because etcdadm typically uses etcd defaults, and by default, etcd does not enable a separate http metrics endpoint.

I think you can use https://godoc.org/github.com/spf13/pflag#FlagSet.Changed to tell between `etcdadm init` and `etcdadm init --listen-metrics-url`. WDYT?

suhrud-kumar:
> > etcdadm init -> no additional http /metrics endpoint
> > etcdadm init --listen-metrics-url -> additional http /metrics endpoint on http://:defaultport
> > etcdadm init --listen-metrics-url http://ip:port -> additional http /metrics endpoint on use requested URL
> 
> I favor this, because etcdadm typically uses etcd defaults, and by default, etcd does not enable a separate http metrics endpoint.

Yes, that is what I wanted to implemented at first. But since this is the only flag that uses this, it could somewhat mess up existing code structure.

> I think you can use https://godoc.org/github.com/spf13/pflag#FlagSet.Changed to tell between `etcdadm init` and `etcdadm init --listen-metrics-url`. WDYT?

pflag would throw an error if there is no argument or no default value for a flag.

[root@ip-172-31-16-230 tmp]# ./etcdadm init --snapshot
Error: flag needs an argument: --snapshot

[root@ip-172-31-16-230 tmp]# ./etcdadm init --listen-metrics-url
Error: flag needs an argument: --listen-metrics-url

So we have to use this https://github.com/spf13/pflag#setting-no-option-default-values-for-flags

So, init function looks like 

```
initCmd.PersistentFlags().Var(util.URLValue{URL: &etcdAdmConfig.AdditionalMetricsURL}, "listen-metrics-url", "URL to expose additional /metrics endpoint without TLS")
initCmd.PersistentFlags().Lookup("listen-metrics-url").NoOptDefVal = apis.DefaultAdditionalMetricsURL()
```

and additional method in apis/config.go pkg

```
// DefaultAdditionalMetricsURL sets additional metrics url using node ip
func DefaultAdditionalMetricsURL() string {
	externalAddress, err := defaultExternalAddress()
	if err != nil {
		log.Fatalf("failed to set default DefaultAdditionalMetricsURL: %s", err)
	}
	url := url.URL{
		Scheme: constants.DefaultAdditionalMetricsScheme,
		Host:   fmt.Sprintf("%s:%d", externalAddress.String(), constants.DefaultAdditionalMetricsPort),
	}
	return url.String()
}
```

Let me know if you are okay with this and I can push it. 

Also, I think I found a bug (?) in pflag NoOptDefVal func, it doesn't work in all cases, will open a issue with them.

suhrud-kumar:
> Also, I think I found a bug (?) in pflag NoOptDefVal func, it doesn't work in all cases, will open a issue with them.

Looks like this is an expected behaviour with pflag - https://github.com/spf13/pflag/issues/134
https://github.com/spf13/pflag#command-line-flag-syntax

So, if we use NoOptDefVal for this param
--listen-metrics-url                                             # valid, sets to default ip
--listen-metrics-url=http://\<node-ip>:1234    # valid, sets to user provided ip
--listen-metrics-url http://\<node-ip>:1234     # invalid, sets to default ip

PR #162, opened 2020-07-25
Add etcd-manager under etcd-manager/ subdirectory

justinsb:
As discussed in office hours, adding etcd-manager in a subdirectory to enable the projects to converge.

PR #171, opened 2020-08-29
Exclude etcd-manager from the style checks for the moment

justinsb:
We can then merge the repos and fix the problems here.


justinsb:
So I think that if we can merge this, I can do a final rebase on #162 and it should be green.  Then I can make the golint / govet fixes here.

cc @dlipovetsky

PR #172, opened 2020-09-02
Temporarily skip etcd-manager gotest

justinsb:
The go.mod means we'll have to run the tests from that subdirectory
(or converge the set of go.mod dependencies).


dlipovetsky:
Makes sense. Thank you!

/lgtm

PR #176, opened 2020-09-28
Always use etcdctl v3 interface

### LINKED ISSUE #173: init: unknown flag: --version
For etcd 3.4.9, etcd does, but etcdctl doesn't support --version flag

```
etcdadm init -l debug
INFO[0000] [install] extracting etcd archive /var/cache/etcdadm/etcd/v3.4.9/etcd-v3.4.9-linux-amd64.tar.gz to /tmp/etcd554743000
INFO[0000] [install] verifying etcd 3.4.9 is installed in /opt/bin/
FATA[0000] [install] Error: command "/opt/bin/etcdctl" failed: "Error: unknown flag: --version\n\nError: unknown flag: --version\n"
# etcdctl --version
Error: unknown flag: --version
```

dlipovetsky:
etcdadm supports only etcd v3, so it is correct to always use the etcdctl v3 interface.

Fixes #173.

### DIFF
--- a/binary/etcd.go
+++ b/binary/etcd.go
@@ -66,11 +66,8 @@ func isEtcdctlInstalled(version, inputDir string) (bool, error) {
 	if !exists {
 		return false, nil
 	}
-	cmdVersionFlag := "--version"
-	if os.Getenv("ETCDCTL_API") == "3" {
-		cmdVersionFlag = "version"
-	}
-	cmd := exec.Command(path, cmdVersionFlag)
+	cmd := exec.Command(path, "version")
+	cmd.Env = append(os.Environ(), "ETCDCTL_API=3")
 	return util.CmdOutputContains(cmd, fmt.Sprintf("etcdctl version: %s", version))
 }
 


binary/etcd.go
```diff
@@ -66,11 +66,8 @@ func isEtcdctlInstalled(version, inputDir string) (bool, error) {
 	if !exists {
 		return false, nil
 	}
-	cmdVersionFlag := "--version"
-	if os.Getenv("ETCDCTL_API") == "3" {
-		cmdVersionFlag = "version"
-	}
-	cmd := exec.Command(path, cmdVersionFlag)
+	cmd := exec.Command(path, "version")
+	cmd.Env = append(os.Environ(), "ETCDCTL_API=3")
```
justinsb, line 70: I'm not sure what happens with a dup env var here (i.e. can we define an env var twice?).

justinsb:
I do wonder what happens if the env var is already defined, but this lgtm - thanks for fixing!

/approve
/lgtm

PR #177, opened 2020-09-28
Set default etcd version to 3.4.9, the default for Kubernetes 1.19

dlipovetsky:
I propose setting the default etcd version to whatever is used in the latest Kubernetes release.

Once this merges, I plan to create a new etcdadm release.

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -20,7 +20,7 @@ import "time"
 
 // Command-line flag defaults
 const (
-	DefaultVersion    = "3.3.8"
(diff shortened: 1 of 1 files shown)


dlipovetsky:
This depends on #176

justinsb:
This lgtm.  I don't know when we should update (when a new k8s is released?) but we can tackle that in future!

/approve
/lgtm

dlipovetsky:
Now that #176 merged:

/test pull-etcdadm-test-init

PR #178, opened 2020-10-13
Check if ETCD is already installed, before attempting to install

ntaylor1781:
In situations where access to the internet is blocked, it may be needed/easiest to install an upstream ETCD package. This creates the situation where ETCD is installed, but etcdadm will fail to move forward because it can't download the binary. We can check if it is installed first, and move onto the old logic if it is not.

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -20,7 +20,7 @@ import "time"
 
 // Command-line flag defaults
 const (
-	DefaultVersion    = "3.3.8"
+	DefaultVersion    = "3.4.9"
 	DefaultInstallDir = "/opt/bin/"
 
 	DefaultReleaseURL     = "https://github.com/coreos/etcd/releases/download"


ntaylor1781:
It does not seem to be registering the CLA as signed. The task says commit missing GitHub user, but I see it as associated.

ntaylor1781:
Attempted reverting my change, and adding it with a fixed author. The CLA shows as authorized. Let me know if there is a desired way to clean this up.

dlipovetsky:
Thanks for the PR @ntaylor1781! Looking.

dlipovetsky:
BTW, to appease the CLA Bot, I think you'll need to remove the initial commit and revert commit and force-push to your GitHub branch.

dlipovetsky:
> In situations where access to the internet is blocked, it may be needed/easiest to install an upstream ETCD package. 

Can you help me understand this use case better, please? By "upstream ETCD package," do you mean a [package maintained by a Linux distro](https://pkgs.org/search/?q=etcd), e.g., an rpm, deb, etc?

etcdadm is designed to support the "air-gapped" use case, but it asks the user to fetch an etcd tarball published on the [etcd GitHub releases page](https://github.com/etcd-io/etcd/releases), and place it into the cache directory.

I agree that, if the _both_ the correct etcd and etcdctl binaries are already installed, then etcdadm should not install them again from the cache. 

However, I think that works only under the assumption that etcdadm "manages" the lifecycle of those binaries.

For example, if you install etcd/etcdctl binaries using a system package, but later run `etcdadm reset`, etcdadm will remove the  binaries, and you won't be able to run `etcdadm init/join` until you install the binaries again.

If etcdadm could delegate installation of the binaries to a system package manager, that would work. But it's not trivial to integrate etcdadm with various system package managers, so I'd like to understand why the "download the tarball and put it in the cache" approach does not work well for you.

ntaylor1781:
By upstream ETCD package, I did mean a Linux distro maintained package. Using the cache directory is valid, and something I initially did as a workaround. The package was an easy way out. It mainly came down to not wanting to have a tarball in the bootstrap/config-management git, and not wanting to stand up a logical place for it. 

I should have read the code for reset, as I didn't know that it cleaned up the binaries on reset. It seemed like this would be a simple change that could allow me to be lazy, but it appears that is not the case. 

A flag could be added to not remove on reset, but I feel like the chance of error is pretty high (a user installs through package manager, but doesn't use the proper flag on reset). 

With this new information it probably makes the most sense to close this.

dlipovetsky:
> The package was an easy way out. It mainly came down to not wanting to have a tarball in the bootstrap/config-management git, and not wanting to stand up a logical place for it.

Can you share how (and when) you fetch the etcd system package?

(Also, do you know about `etcdadm download`, which populates the cache without installing etcd?)

ntaylor1781:
The process I was working on was to install etcd from an internal mirrored repo, and then have automation handle the etcdadm commands to start. The idea being that this change would allow it to see that it already existed, and not fail out.  

I am aware of `etcdadm download`, and used it to create the cache directory structure (as the download failed to get to the internet). 

A solution I can use would be to package the cache directory, and etcdadm, in an rpm. I can then easily install that from the internal mirrors. I initially added the change as it seemed like it would have no affect, and make it so that I didn't have to make an RPM (or provide some other system for grabbing the tar).

dlipovetsky:
> A solution I can use would be to package the cache directory, and etcdadm, in an rpm

That would work.

If you don't need an rpm to get etcdadm on the machine, then it might be easier to create an cache-only rpm, e.g., by running `etcdadm download` in the rpm build root.

dlipovetsky:
It's worth contemplating making install/uninstall _opt-in_. Since etcdadm already _verifies_ that the binaries are of the right version, I think it should be safe to delegate install/uninstall to something else.

```shell
etcdadm install
etcdadm init/join

..

etcdadm reset
etcdadm uninstall
```

Alternatively,

```shell
etcdadm init --install
etcdadm reset --uninstall
```

Do you want to join the next etcdadm office hours to talk about the trade-offs?

ntaylor1781:
Yea definitely. We could also look at implementing the phase concept kubeadm uses. This could be useful to run only specific steps, but would also make it easy to add a `--skip-phases` option to init and reset as well.

dlipovetsky:
@ntaylor1781 Could you file an issue for phases, please? If you think phases would meet your use case, we might be able to close this PR.

ntaylor1781:
I created #179 for tracking. This is a busy time of year, but I will see what I can do with it, if someone doesn't beat me to it. Would you like me to close this PR?

ntaylor1781:
Well guess cleaning up my fork to look at phases closed it anyone.

PR #187, opened 2021-01-12
etcd-manager/dev: move script interpreter to the first line

zhijianli88:
Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>

### DIFF
--- a/etcd-manager/dev/add-symlink.sh
+++ b/etcd-manager/dev/add-symlink.sh
@@ -1,3 +1,4 @@
+#!/bin/bash -e
 # Copyright 2020 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


dlipovetsky:
Can you please describe the issue you found that this fixes?

etcd-manager/dev/add-symlink.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash -e
```
dlipovetsky, line 1: `-e` will cause the script to exit on error. Why is this flag used here, but not in the other scripts?

zhijianli88: Looks -e was here since the initial version, and i just want to fix the interpreter issue first in this pr

zhijianli88:
Only the interpreter appears in the first line makes sense

justinsb:
Thanks @zhijianli88 

/approve
/lgtm
/ok-to-test

PR #194, opened 2021-03-21
Build & stage etcd-manager also

justinsb:
Part of moving canonical location of etcd-manager to this repo.

PR #197, opened 2021-03-21
Use go.mod for verify-spelling

justinsb:
Mirrors the approach used in kubernetes/kubernetes, avoids the
spurious failure.


justinsb:
Added the #196 commit as well because it looks like both #197 and #196 are needed to get tests working.  I think this was introduced by go 1.16.

/assign @dlipovetsky

justinsb:
I'm guessing the dependency order for these PRs is #197 (which includes #196) then #195, then #194

PR #198, opened 2021-03-24
etcd-manager - Upgrade AWS SDK

rifelpet:
We expanded the AWS regions and zones that the kops periodic jobs run in and I noticed that some jobs are failing because ap-northeast-3 isn't recognized.

This updates the AWS SDK to properly recognize ap-northeast-3.

[example job](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-k8s-latest/1374641041314942976) and [etcd-manager logs](https://storage.googleapis.com/kubernetes-jenkins/logs/e2e-kops-aws-k8s-latest/1374641041314942976/artifacts/13.208.181.237/etcd.log)

As an aside, I noticed a few build issues, mostly related to the repo move:
* etcdadm's .gitignore ignores `vendor` which etcd-manager has checked in. We should either migrate off of using vendor or update the .gitignore pattern to only match the repo's top level vendor directory (or any others that _should_ be ignored)
* the `tools/deb-tools` go submodule is still using the kope.io path and should be updated. EDIT: https://github.com/kubernetes-sigs/etcdadm/pull/199
* `tools/deb-tools` also doesn't use the etcd-manager directory's .bazelversion file and so bazelisk defaulted to 4.X during `make vendor`. we should probably copy or link the `etcd-manager/.bazelversion` file


justinsb:
Thanks!

/approve
/lgtm

PR #200, opened 2021-03-26
Install bazel in staging build

justinsb:
A little inefficient, but means that we can control our bazel version
and not rely on the upstream image.


justinsb:
I updated this to use bazelisk, so that we work the same way as our github actions build.

PR #202, opened 2021-04-04
staging artifact upload: avoid double slash

justinsb:
A double slash is preserved in GCS, so make sure that we don't introduce it!

Also fix source location for etcd-manager binaries.


justinsb:
/retest

500 Internal Server Error fetching debian packages

justinsb:
/retest

Another 500 Internal Server Error fetching debian packages

PR #205, opened 2021-04-22
Remove support for etcd v2

hakman:
Tested with kOps in HA setup and seems fine:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/11341/pull-kops-e2e-k8s-containerd-ha/1387265704251625472

/cc @justinsb


etcd-manager/pkg/etcd/etcdprocess.go
```diff
@@ -236,12 +236,6 @@ func (p *etcdProcess) Start() error {
 		env["ETCD_INITIAL_CLUSTER_STATE"] = "existing"
 	}
 
-	// For etcd3, we disable the etcd2 endpoint
-	// The etcd2 endpoint runs a weird "second copy" of etcd
-	if !strings.HasPrefix(p.EtcdVersion, "2.") {
-		env["ETCD_ENABLE_V2"] = "false"
```
johngmyers, line 242: Don't we want to still disable the etcd2 endpoint on etcd3? Shouldn't this code be unconditional, not removed?

johngmyers (changes requested):

etcd-manager/pkg/etcd/etcdprocess.go
hakman: `false` is the default for some time:
https://etcd.io/docs/v3.5/op-guide/configuration/#--enable-v2

hakman: Done. I agree that should be set at least while we support etcd <= 3.3.

hakman:
Thanks @johngmyers!

PR #214, opened 2021-04-29
Makefile: Update go version from 1.12 to 1.16

dlipovetsky:
Version 1.12 is no longer supported.


mrajashree:
lgtm 👍

hakman:
Maybe should also change in go.mod in a future commit:
https://github.com/kubernetes-sigs/etcdadm/blob/c09b9e2d7cca67372328772f483c1f3249f25247/go.mod#L3

/lgtm

PR #236, opened 2021-07-08
[WIP] Sync bundled etcd versions with k/k

hakman:
xRef:
* [k/k/cluster/images/etcd/Makefile#L18](https://github.com/kubernetes/kubernetes/blob/e974bcd471fa562a000ca9780a479c01b0dfe84f/cluster/images/etcd/Makefile#L18)
* [k/k/cmd/kubeadm/app/constants/constants.go#L461-L471](https://github.com/kubernetes/kubernetes/blob/d7123a65248e25b86018ba8220b671cd483d6797/cmd/kubeadm/app/constants/constants.go#L461-L471)


hakman:
With this change, the time it take to run the Upgrade/Downgrade test decreases from 48 min to 12 min.

hakman:
Replaced by #244.

PR #237, opened 2021-07-23
add custom bind addresss support to init

rockmenjack:
this pull request adds `--bind-address` to specify listen address generated in `etcd.env`

### DIFF
--- a/apis/config.go
+++ b/apis/config.go
@@ -257,7 +257,7 @@ func DefaultClientURLs(cfg *EtcdAdmConfig) error {
 
 // DefaultInitialAdvertisePeerURLs TODO: add description
 func DefaultInitialAdvertisePeerURLs(cfg *EtcdAdmConfig) error {
-	externalAddress, err := defaultExternalAddress(cfg.BindAddr)
+	externalAddress, err := ensureValidExternalAddress(cfg.BindAddr)
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


rockmenjack:
CLA signed

dlipovetsky:
Thank you @rockmenjack!

/ok-to-test

apis/config.go
```diff
@@ -301,10 +304,37 @@ func DefaultAdvertiseClientURLs(cfg *EtcdAdmConfig) error {
 }
 
 // Returns the address associated with the host's default interface.
```
dlipovetsky, line 306: I think the logic in the function is sound, but this function no longer does what its comment says it does.

```suggestion
// Returns the user-defined address; if that is undefined, returns the address associated with the host's default interface.
```

(I think refactoring this code would be nice, but not required for this PR.)

rockmenjack: pull request updated

dlipovetsky:
@rockmenjack, sorry for the delay, @justinsb and I are looking at the PR at the office hours this morning, and we see that `defaultExternalAddress` function actually does validation of the user-defined address, and returns an error if the validation fails. That code looks like it should belong in a different function. Would you be ok if I added a commit to your PR, or would you like to try doing it yourself?

rockmenjack:
How about just rename the function and update its comment to reflect the changed behavior? (see above commit)

PR #238, opened 2021-07-27
Update OpenStack Client to Use Volume V3 API

iGene:
OpenStack Volume V2 API is deprecated since Pike and newer version of
OpenStack deployment may not include volume V2 API endpoints. While
volume V3 API is compatible with V2, we can just switch the client to
use V3 API.
This commit switches etcd-manager to use OpenStack Volume V3 API.


iGene:
I would appreciate if any of the reviewers can review this patch.
Thanks in advance!

hakman:
@zetaab Any chance you can take a look at this?

PR #240, opened 2021-08-30
Run staticcheck using new versioned go-run syntax

justinsb:
This requires go 1.17, but gets us out of a catch-22 where the gofmt
test is failing with the latest go version (harder to fix as
configured in the test-infra job), but moving to go 1.17 syntax would
break compliation for go 1.16 (which is the most common version right
now).

This requires go 1.17 _only_ for these two makefile targets.


neolit123 (commented):
/lgtm

PR #248, opened 2021-11-01
Create test for backup and restore functionality

justinsb:
We create a simple test that we can backup and restore, and that a
(single) key-value pair persists.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
 
     - name: Test etcd-manager
       run: |
-        "${GITHUB_WORKSPACE}/bin/bazelisk" test --test_output=streamed -- //... -//test/integration/upgradedowngrade:all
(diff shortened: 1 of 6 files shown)


hakman (approved):

etcd-manager/test/integration/BUILD.bazel
```diff
@@ -31,6 +32,41 @@ go_test(
...
+        "//pkg/etcd",
+        "//pkg/etcdversions",
+        "//test/integration/harness",
+        "//vendor/k8s.io/klog/v2:klog",
+    ],
+)
+
+go_test(
+    name = "integration_backuprestore_test",
+    size = "enormous",  # Takes more than 900 seconds to run on travis
```
hakman, line 45: Would it make sense to move the test to a separate job similar to what we do for with `test-upgradedowngrade`?
https://github.com/kubernetes-sigs/etcdadm/blob/40001ec166bac2cf049ac2dd5ef63c68394aa70a/.github/workflows/main.yml#L41

hakman (changes requested):

PR #252, opened 2021-11-01
Use unixs for "restore source" endpoint

justinsb:
When restoring a backup for both etcd and etcd-events, I observed a
port-conflict on port 8002/8003.  Switching to unixs:// endpoints (unix
domain sockets) is a little more secure, and avoids the port conflict.

Due to https://github.com/etcd-io/etcd/issues/12450 the syntax is
slightly awkward - we have to make it look like a host and port.

### DIFF
--- a/etcd-manager/pkg/etcd/restore.go
+++ b/etcd-manager/pkg/etcd/restore.go
@@ -110,18 +110,12 @@ func RunEtcdFromBackup(backupStore backup.Store, backupName string, basedir stri
 		return nil, err
 	}
 
-	isV2 := false
 	if strings.HasPrefix(backupInfo.EtcdVersion, "2.") {
-		isV2 = true
+		return nil, fmt.Errorf("this version of etcd-manager does not support etcd v2")
 	}
 
-	var downloadFile string
-	if isV2 {
-		downloadFile = filepath.Join(basedir, "download", "backup.tar.gz")
-	} else {
-		// V3 requires that data dir not exist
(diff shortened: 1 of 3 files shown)


justinsb:
WIP because - although this does seem to work on all the versions of etcd (verified by backup/restore testing) - I don't know if we want to only use it for e.g. versions greater than 3.x.  On the other hand, we don't know of any problems, testing doesn't show any, and because this is purely local we can change this easily should any problems arise.

justinsb:
After more though, I think we should just introduce this for all versions - the underlying bug where the ports can conflict has existed for all versions, and we haven't observed any problems in testing.

etcd-manager/pkg/etcd/restore.go
```diff
@@ -143,17 +143,37 @@ func RunEtcdFromBackup(backupStore backup.Store, backupName string, basedir stri
...
 	myNode := &protoetcd.EtcdNode{
-		Name:       myNodeName,
-		ClientUrls: []string{clientUrl},
-		PeerUrls:   []string{peerUrl},
+		Name: myNodeName,
 	}
+
+	currentDir := ""
+
+	if !isV2 {
```
hakman, line 153: I thought I removed all support for V2. Should we remove this part also?

justinsb: Good point - removed all V2 support from this function - it wouldn't work anyway as we don't ship etcd2!

hakman: :)

PR #253, opened 2021-11-01
etcd-manager - Allow all 3.X upgrades of one minor version

rifelpet:
This is a bit more future-proof and lower maintenance. This should cover all 3.X upgrades of up to one minor version.

ref: https://github.com/kubernetes/kops/issues/12662


justinsb:
Yes, this seems like a good idea.  We could also enforce this via tests, but as we're working towards supporting etcd versions running in a separate container/pod, this approach is more sustainable.

Thanks @rifelpet 

/approve
/lgtm

PR #254, opened 2021-11-15
Install bazelisk for bazel and bazelisk

justinsb:
We were only using it for bazel invocations, but as we adopt more
bazelisk we would like to be able to invoke bazelisk explictly.


rifelpet (approved):
/lgtm

PR #255, opened 2021-11-15
Split staging builds for etcdadm and etcd-manager

justinsb:
Because the versioning schemes are different (for now), and we want to
avoid rebuilds/reuploads to the same location.


justinsb:
Supporting test-infra change is here: https://github.com/kubernetes/test-infra/pull/24354

justinsb:
Upstream merged; cancelling hold

/hold cancel

PR #257, opened 2021-11-16
Update go.mod to v1.17

hakman:


johngmyers:
Looks like `hack/verify-deps.sh` doesn't verify the etcd-manager deps or vendoring. Still,
/lgtm

PR #258, opened 2021-11-16
Fix release of built assets for etcd-manager

hakman:
Ref: https://testgrid.k8s.io/sig-k8s-infra-gcb#etcdadm-postsubmit-push-to-staging


rifelpet (approved):
/lgtm

PR #259, opened 2021-11-16
staging builds: compute our own version

justinsb:
The version that is passed to us includes a date prefix, so we need to
compute our own tag.

Do this (initially) for etcd-manager only.


justinsb:
Thanks for approving @rifelpet - did you intend not to give lgtm?

rifelpet:
oops
/lgtm

PR #261, opened 2021-11-24
Replace etcd v3.5.0 with etcd v3.5.1

hakman:
xRefs:
* https://github.com/etcd-io/etcd/releases/tag/v3.5.1
* https://github.com/kubernetes/kubernetes/issues/106589

/cc @rifelpet @olemarkus


hakman:
@justinsb Would it be ok to do a new release with this change?

PR #263, opened 2021-11-24
Remove the service arg from InitSystem

justinsb:
It was only passed a constant value, and arguably the wrong value if
the user overrides the UnitFile.

Fixing this moves us closer to reuse of the InitSystem.


rifelpet (approved):
/lgtm

PR #270, opened 2021-11-28
Add support for listening to IPv6

olemarkus:
If instances are using Resource Based Naming, we are guaranteed to have ipv6 addresses and we use them between etcd-manager instances.

This will also make etcd-manager work in ipv6-native/only subnets.

This PR also makes it possible for etcd-manager to create an etcd cluster listening on ipv6, but the default remains listening on ipv4.

This PR is tested in https://github.com/kubernetes/kops/pull/12855

### DIFF
--- a/etcd-manager/pkg/volumes/aws/volumes.go
+++ b/etcd-manager/pkg/volumes/aws/volumes.go
@@ -103,11 +103,19 @@ func NewAWSVolumes(clusterName string, volumeTags []string, nameTag string) (*AW
 		return nil, fmt.Errorf("error querying ec2 metadata service (for instance-id): %v", err)
 	}
 
-	a.localIP, err = a.metadata.GetMetadata("local-ipv4")
-	if err != nil {
-		return nil, fmt.Errorf("error querying ec2 metadata service (for local-ipv4): %v", err)
+	hostnameBytes, err := a.metadata.GetMetadata("local-hostname")
+	hostname := string(hostnameBytes)
+	if strings.HasPrefix(hostname, "i-") {
+		a.localIP, err = a.metadata.GetMetadata("ipv6")
+		if err != nil {
+			return nil, fmt.Errorf("error querying ec2 metadata service (ipv6): %v", err)
+		}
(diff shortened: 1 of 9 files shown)


etcd-manager/pkg/volumes/aws/volumes.go
```diff
@@ -98,16 +99,29 @@ func NewAWSVolumes(clusterName string, volumeTags []string, nameTag string) (*AW
 		return nil, fmt.Errorf("error querying ec2 metadata service (for az): %v", err)
 	}
 
-	a.instanceID, err = a.metadata.GetMetadata("instance-id")
+	hostname, err := os.Hostname()
```
hakman, line 102: I think getting this from metadata would be more reliable.

olemarkus: I debated this a bit ... metadata has no notion of a primary IPv6 address. So which one would be the right one in the metadata. Always assume index 0's address? lowest index if index 0 has no IP? In CCM there is some dead code for ipv6 from metadata and it is not that nice.

hakman: Meant the hostname.

olemarkus: If we want to get rid of hostname override for RBN in kOps, we need to rely on hostname working. If not, kubelet/kube-proxy will also fail at this.

hakman: Yes, but os.Hostname ran in a container doesn't seem like the best thing. Using metadata makes is OS independent.

olemarkus: That's a good point. It works for kOps because of hostnetworking and such. Without that, this will probably fail. But then, metadata is maybe not available either.

Right now, there is a flag for passing in listen address, but it is ignored, and etcd-manager will always look up. But yeah, I can switch this to looking up this from metadata.

hakman (approved):

PR #273, opened 2021-12-02
Finish cert renewal

ntaylor1781:
This is continuing the work from [#147](https://github.com/kubernetes-sigs/etcdadm/pull/147) done by @pytimer.
If there is a different method desired for doing this, please let me know. I would find this feature super helpful, 
and found the time to dig into it. 

Last concern was about an issues with CSR generation. This was due to trying to write the
CSR and key to the certificate directory. When calling TryLoadCSRAndKeyFromDisk it registered
the key existed, and tried to load both. This caused the error stating it couldn't load the
CSR.

As with kubeadm, this needed a directory to output the csr specifically. I added a new argument
of csr-dir (with a default of the local directory). I split the RenewUsingLocalCA function into two
functions. One for the CSRs, and one for the full certificates. The new RenewCSRUsingLocalCA
function uses the new csr-dir path to output the CSRs. This ensures the key, from the cert being
used, is not overwritten/cause the process to fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,22 @@ On the machine being removed, run
 etcdadm reset
 ```
 
+### Renewal certificates
+
+1. Ensure the CA certificate and key already exists on the machine.
+
+If not exists, Copy the CA certificate and key from other machine in the cluster.
+
+```
+rsync -avR /etc/etcd/pki/ca.* <Member IP address>:/
+```
+
+2. run command to renew certficates via CA certificate and key.
+
+```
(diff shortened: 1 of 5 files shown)


ntaylor1781:
@justinsb  or @rifelpet is there anything you need from me?

rifelpet:
I'm not familiar with this are of code unfortunately
/uncc @rifelpet

ntaylor1781:
@dlipovetsky can I get a review?

PR #275, opened 2021-12-08
Update k8s dependencies to v1.23.1

hakman:
/cc @rifelpet @olemarkus


etcd-manager/.gitignore
```diff
@@ -20,4 +20,4 @@ bazel-testlogs
 dist/
 
 # Don't ignore vendoring
-!vendor/**
+!vendor/
```
rifelpet, line 23: Why is this changing?

rifelpet (approved):

olemarkus:
Sorry for missing this one. Want to bump to 1.23.3?

hakman:
> Sorry for missing this one. Want to bump to 1.23.3?

Thanks. Let's keep it to 1.23.1, same as kOps for now. There's not much difference in using 1.23.3 anyway.

olemarkus (approved):

etcd-manager/.gitignore
justinsb: lgtm - I would have thought they would be the same based on the docs, but I can confirm this does (correctly) include the version directory.

I _think_ it’s because we need to match the parent .gitignore (or maybe the issue is the vendor directory _itself_ is not toggled by `vendor/**`, but anyway this now lgtm.)

PR #279, opened 2022-01-25
Add back etcd 3.5.0

olemarkus:
#261 removed the 3.5.0 binary, which breaks at least all kops upgrades from 1.22.{0,1,2} to 1.22.3, which already uses 3.5.0. The reason being that etcd-manager will first start etcd using the current version, and only then start etcd with the new version.

/cc @zetaab @hakman

xref: https://github.com/kubernetes/kops/issues/13118


olemarkus:
Not sure why it cannot clean up the files of that node. I cannot reproduce this locally.
/retest

olemarkus:
Ah doh. it is of course not reacting to retest ...

rifelpet (approved):
/lgtm

hakman:
It will not merge until the test passes...

rifelpet (approved):

PR #281, opened 2022-01-28
staging build: increase timeout to 30 minutes

justinsb:
We are now seeing timeouts with the 10 minute default; raise to 30
minutes.  We aren't running the tests, so 30 minutes should be plenty.


justinsb:
We are just on the cusp of the 10 minute threshold; we "got lucky" on the most recent run and it ran in 9m52s

rifelpet (approved):

PR #292, opened 2022-02-03
Test for nil in to not found check

olemarkus:
/cc @hakman

### DIFF
--- a/etcd-manager/pkg/volumes/aws/volumes.go
+++ b/etcd-manager/pkg/volumes/aws/volumes.go
@@ -103,7 +103,13 @@ func NewAWSVolumes(clusterName string, volumeTags []string, nameTag string) (*AW
 		return nil, fmt.Errorf("error querying ec2 metadata service (for instance-id): %v", err)
 	}
 
+	a.ec2 = ec2.New(s, config.WithRegion(region))
 	a.localIP, err = a.metadata.GetMetadata("ipv6")
+
+	// If we have an IPv6 address, return
+	if err == nil {
+		return a, nil
+	}
 	awsErr, isAWSErr := err.(awserr.RequestFailure)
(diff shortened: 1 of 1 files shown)


etcd-manager/pkg/volumes/aws/volumes.go
```diff
@@ -104,8 +104,7 @@ func NewAWSVolumes(clusterName string, volumeTags []string, nameTag string) (*AW
 	}
 
 	a.localIP, err = a.metadata.GetMetadata("ipv6")
-	awsErr, isAWSErr := err.(awserr.RequestFailure)
-	if !isAWSErr || awsErr.StatusCode() != 404 {
+	if !isNotFoundError(err) {
```
hakman, line 107: Also, maybe `a.ec2 = ec2.New(s, config.WithRegion(region))` can be put before this bloc to not duplicate the code before return.

```suggestion
	if err == nil {
		a.ec2 = ec2.New(s, config.WithRegion(region))
 		return a, nil
 	}
 	
	awsErr, isAWSErr := err.(awserr.RequestFailure)
	if !isAWSErr || awsErr.StatusCode() != 404 {
```

hakman (changes requested):

etcd-manager/pkg/volumes/aws/volumes.go
olemarkus: Won't this duplicate the logic? This is doing one `return a` for ipv6 and then a `return a` for ipv4.

hakman: I assume that if we have ipv6, we want to return that one. As is in the PR, it will return ipv4 for dual-stack too.

olemarkus: You are absolutely correct. I missed that.

hakman (approved):

PR #297, opened 2022-03-02
Allow using multiple endpoints for etcd client

mrajashree:
The etcdadm init command creates a single member etcd cluster. Other
members can be added to the cluster using the etcdadm join command with
the endpoint of the first member.
If this first member gets deleted, we need to decide on a different endpoint
to be used for the join command for future members. Instead, once the
cluster is created, we should be able to use endpoints of all members with the
join command.
This endpoint only gets used to create an etcd client for listing/adding members.
This commit modifies the ClientForEndpoint method to use all
endpoints that are passed in. This way the etcdadm join command can accept
a comma separated list of etcd members.

### DIFF
--- a/cmd/join.go
+++ b/cmd/join.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strings"
 
 	log "sigs.k8s.io/etcdadm/pkg/logrus"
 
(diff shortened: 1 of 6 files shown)


cmd/join.go
```diff
@@ -84,11 +85,13 @@ func newJoinRunner() *runner {
 }
 
 func joinPhasesSetup(cmd *cobra.Command, args []string) (*phaseInput, error) {
-	endpoint := args[0]
-	if _, err := url.Parse(endpoint); err != nil {
-		return nil, fmt.Errorf("endpoint %q must be a valid URL: %s", endpoint, err)
+	endpoints := strings.Split(args[0], ",")
+	for _, endpoint := range endpoints {
+		if _, err := url.ParseRequestURI(endpoint); err != nil {
+			return nil, fmt.Errorf("endpoint %q must be a valid URL: %s", endpoint, err)
```
justinsb, line 91: Nit: I think we should use %w (`fmt.Errorf("...: %w", ..., err)`) for any errors now; at some stage we'll have to go back and fix the existing invocations, but we should probably start using the chaining form as it's a few go versions old at this point.

Not a blocker as we have tons of %v and %s already though!

justinsb:
This LGTM - it introduces multiple endpoints without breaking backwards compatability.

/lgtm

@dlipovetsky are you OK for me to start approving PRs in etcdadm itself (vs in etcd-manager)?

PR #301, opened 2022-03-18
Added flags for client-port and peer port

rh01:


rh01:
Missing CLA, close PR first, will recreate a new PR.

PR #302, opened 2022-03-18
 Added flags for client-port and peer port

rh01:
Added flags for client-port and peer port


certs/pkiutil/pki_helpers.go
```diff
@@ -34,12 +34,13 @@ import (
 	"path/filepath"
 	"time"
```
justinsb, line 36: Thanks for doing this .. as a nit though, we should probably combine this into the block above (with other "stdlib" imports)

apis/config.go
```diff
@@ -208,12 +212,7 @@ func SetDownloadDynamicDefaults(cfg *EtcdAdmConfig) error {
 	return setDynamicDefaults(cfg)
 }
 
-// SetDefaults sets configuration values defined at build time
-func SetDefaults(cfg *EtcdAdmConfig) {
-	cfg.UnitFile = constants.UnitFile
-	cfg.EnvironmentFile = constants.EnvironmentFile
-	cfg.EtcdctlEnvFile = constants.EtcdctlEnvFile
```
justinsb, line 215: This pattern has evolved because it supports defaults in both flags and when used in code.  IMO we should set the defaults in SetDefaults, and then the flags should reference the values in cfg, but we don't _always_ do that.

But if we're following the pattern, then we would set ClientPort and PeerPort to constants.DefaultClientPort / constants.DefaultPeerPort here

apis/config.go
```diff
@@ -58,7 +58,11 @@ type EtcdAdmConfig struct {
 	CacheDir   string
 	PodSpecDir string
 
-	BindAddr string
+	BindAddr   string
+	ClientPort int
+	PeerPort   int
+
+	CertValid int
```
justinsb, line 65: `CertValidity time.Duration` or `CertLifetime time.Duration` might be clearer here - I originally thought that we were turning off cert validation, and also by using time.Duration we avoid having to know that it is in years.

cmd/init.go
```diff
@@ -43,6 +43,12 @@ func init() {
 	initCmd.PersistentFlags().StringVar(&etcdAdmConfig.ReleaseURL, "release-url", constants.DefaultReleaseURL, "URL used to download etcd")
 	initCmd.PersistentFlags().StringVar(&etcdAdmConfig.CertificatesDir, "certs-dir", constants.DefaultCertificateDir, "certificates directory")
 	initCmd.PersistentFlags().StringVar(&etcdAdmConfig.BindAddr, "bind-address", "", "etcd bind address")
+	initCmd.PersistentFlags().StringVar(&etcdAdmConfig.UnitFile, "unit-file", "/etc/systemd/system/etcd.service", "system unit file")
```
justinsb, line 46: So IMO we should use either `constants.Foo` or the etcdAdmConfig value (which would have been set in SetDefaults).  It looks like we're doing the first form, though I personally prefer the second form.  The second form would be e.g.

`initCmd.PersistentFlags().IntVar(&etcdAdmConfig.PeerPort, "peer-port", etcdAdmConfig.PeerPort, "etcd port listening from peer etcd instance")`

cmd/init.go
```diff
@@ -86,7 +92,6 @@ func newInitRunner() *runner {
 }
 
 func initPhasesSetup(_ *cobra.Command, _ []string) (*phaseInput, error) {
-	apis.SetDefaults(&etcdAdmConfig)
```
justinsb, line 89: If we are following the SetDefaults pattern, we should keep this

constants/constants.go
```diff
@@ -27,6 +27,7 @@ const (
 	DefaultImageRepository = "quay.io/coreos/etcd"
 	DefaultCertificateDir  = "/etc/etcd/pki"
 
+	// TODO: replace default name
```
justinsb, line 30: That's going to be a breaking change, so it might not be easy.  But what are you thinking of here?  Supporting multiple instances of etcd, or something else?

constants/constants.go
```diff
@@ -142,6 +143,8 @@ ETCD_STRICT_RECONFIG_CHECK=true
 
 	EtcdctlEnvFileTemplate = `export ETCDCTL_API=3
 
+export ETCDCTL_ENDPOINTS={{ .ListenClientURLs.String }}
```
justinsb, line 146: Interesting change - it looks right, but I'm wondering how we were passing ListenClientURLs previously?

justinsb:
Thanks for this @rh01 - I took a first pass and it looks like a great contribution - these options should be configurable.  A few comments on the options pattern, which can allow reuse of the logic beyond just flags - let me know what you think.  I'm guessing you're working towards to support multiple instances of etcd on one machine?

It would be great to squash the commits, and if you wanted to the peer/client port changes could be separated from the cert-validity lifetime changes.  But not a huge deal, they're sort of generally related by "more configuration"!

Are you able to sign the CLA though?  It can be a bit fiddly, but my guess is that the email address on your commits is not linked  to your github account.

PR #303, opened 2022-04-13
etcd-manager - add support for etcd v3.5.3

rifelpet:
https://github.com/etcd-io/etcd/releases/tag/v3.5.3


etcd-manager/pkg/etcdversions/mappings.go
```diff
@@ -116,7 +118,11 @@ func EtcdVersionForAdoption(fromVersion string) string {
 			return Version_3_4_13
 		}
 	case "3.5":
-		return Version_3_5_1
+		if fromSemver.Patch <= 1 {
+			return Version_3_5_1
+		} else {
+			return Version_3_5_3
+		}
```
rifelpet, line 125: I wasn't sure what to do here exactly, whether we want to allow users to keep using 3.5.1 or force them to 3.5.3 (I think that's what this logic handles?)

hakman: I vote for force. This was the reason we don't have 3.5.0 either.

rifelpet: I've updated this to force 3.5.3 in both functions similar to what we did for 3.5.0 -> 3.5.1

hakman (approved):

justinsb:
Thanks for doing this @rifelpet !

PR #304, opened 2022-04-16
Set ETCD_EXPERIMENTAL_INITIAL_CORRUPT_CHECK for etcd 3.5.x

justinsb:
This is the recomended mitigation for the 3.5 series, and is planned
to be the default in 3.6

### DIFF
--- a/etcd-manager/pkg/etcd/etcdprocess.go
+++ b/etcd-manager/pkg/etcd/etcdprocess.go
@@ -32,6 +32,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/blang/semver/v4"
 	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog/v2"
 	protoetcd "sigs.k8s.io/etcdadm/etcd-manager/pkg/apis/etcd"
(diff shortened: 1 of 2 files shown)


etcd-manager/pkg/etcd/etcdprocess.go
```diff
@@ -278,6 +279,17 @@ func (p *etcdProcess) Start() error {
 		klog.Warningf("using insecure configuration for etcd clients")
 	}
 
+	// etcd 3.5 had some corruption issues and recommends this setting,
+	// which is planned as the default in 3.6
+	version, err := semver.ParseTolerant(p.EtcdVersion)
+	if err != nil {
+		klog.Warningf("error parsing version %q: %v", p.EtcdVersion, err)
+	}
```
hakman, line 287: Nit: In the past, this used to be done with just `HasPrefix`, probably to avoid the error handling, without exiting.
https://github.com/kubernetes-sigs/etcdadm/blob/11636446c1b71bde18a055180de1099b886f08d3/etcd-manager/pkg/etcd/etcdprocess.go#L241-L243

hakman (approved):

etcd-manager/pkg/etcd/etcdprocess.go
justinsb: Ah good point ... do you think it's a blocker?  I started the release process...

hakman: Surely not a blocker :)

PR #317, opened 2022-06-14
OpenStack: Filter fixed IP addresses with CIDR notation

ederst:
Currently, `etcd-manager` returns the first found fixed IP address on VMs with multiple IP addresses in OpenStack.

However, there are multiple implications, for example:
* The first found IP address might not be the right one
* The first found IP address can vary everytime `etcd-manager` is executed due to inconsistent sorting of the results returned by the OpenStack compute API
* ETCD members can choose IP addresses located in different subnets which can lead to a non-functioning cluster 

To mitigate this, let users specify a filter with CIDR notation so that instead of the first found fixed IP address the first IP address matching the CIDR notation will be returned.

When no filter is specified `etcd-manager` will still return the first found fixed IP address.

### DIFF
--- a/etcd-manager/cmd/etcd-manager/main.go
+++ b/etcd-manager/cmd/etcd-manager/main.go
@@ -101,6 +101,8 @@ func main() {
 	var volumeTags stringSliceFlag
 	flag.Var(&volumeTags, "volume-tag", "tag which volume is required to have")
 
+	flag.StringVar(&o.NetworkCIDR, "network-cidr", o.NetworkCIDR, "filter for a specific IP address by network CIDR (OpenStack only)")
+
 	flag.Parse()
 
 	o.VolumeTags = volumeTags
@@ -162,6 +164,9 @@ type EtcdManagerOptions struct {
 
 	// EtcdManagerMetricsPort allows exposing statistics from etcd-manager
 	EtcdManagerMetricsPort int
+
+	// NetworkCIDR allows filtering for a specific IP address by network CIDR (OpenStack only)
+	NetworkCIDR string
 }
 
 // InitDefaults populates the default flag values
@@ -200,6 +205,19 @@ func RunEtcdManager(o *EtcdManagerOptions) error {
 
 	if o.BackupStorePath == "" {
 		return fmt.Errorf("backup-store is required")
+	}
+
+	var networkCIDR *net.IPNet
+	if o.NetworkCIDR != "" {
+		if o.VolumeProviderID != "openstack" {
+			return fmt.Errorf("network-cidr is only supported with provider 'openstack'")
+		}
+
+		var err error
+		_, networkCIDR, err = net.ParseCIDR(o.NetworkCIDR)
+		if err != nil {
+			return fmt.Errorf("parsing network-cidr: %w", err)
+		}
 	}
 
 	backupInterval, err := time.ParseDuration(o.BackupInterval)
@@ -244,7 +262,7 @@ func RunEtcdManager(o *EtcdManagerOptions) error {
 			discoveryProvider = gceVolumeProvider
 
 		case "openstack":
-			osVolumeProvider, err := openstack.NewOpenstackVolumes(o.ClusterName, o.VolumeTags, o.NameTag)
+			osVolumeProvider, err := openstack.NewOpenstackVolumes(o.ClusterName, o.VolumeTags, o.NameTag, networkCIDR)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "%v\n", err)
 				os.Exit(1)
--- a/etcd-manager/pkg/volumes/openstack/util.go
+++ b/etcd-manager/pkg/volumes/openstack/util.go
@@ -18,6 +18,7 @@ package openstack
 
 import (
 	"fmt"
+	"net"
 )
 
 const (
@@ -26,7 +27,7 @@ const (
 	openstackAddress        = "addr"
 )
 
-func GetServerFixedIP(addrs map[string]interface{}, name string) (poolAddress string, err error) {
+func GetServerFixedIP(addrs map[string]interface{}, name string, networkCIDR *net.IPNet) (poolAddress string, err error) {
 	for _, address := range addrs {
 		if addresses, ok := address.([]interface{}); ok {
 			for _, addr := range addresses {
@@ -34,6 +35,9 @@ func GetServerFixedIP(addrs map[string]interface{}, name string) (poolAddress st
 				if addrType, ok := addrMap[openstackExternalIPType]; ok && addrType == openstackAddressFixed {
 					if fixedIP, ok := addrMap[openstackAddress]; ok {
 						if fixedIPStr, ok := fixedIP.(string); ok {
+							if networkCIDR != nil && !networkCIDR.Contains(net.ParseIP(fixedIPStr)) {
+								continue
+							}
 							return fixedIPStr, nil
 						}
 					}
--- a/etcd-manager/cmd/etcd-manager/BUILD.bazel
+++ b/etcd-manager/cmd/etcd-manager/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
 
 go_library(
     name = "etcd-manager_lib",
@@ -39,4 +39,10 @@ go_binary(
     name = "etcd-manager",
     embed = [":etcd-manager_lib"],
     visibility = ["//visibility:public"],
+)
+
+go_test(
+    name = "etcd-manager_test",
+    srcs = ["main_test.go"],
+    embed = [":etcd-manager_lib"],
 )
--- a/etcd-manager/cmd/etcd-manager/main_test.go
+++ b/etcd-manager/cmd/etcd-manager/main_test.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"net"
+	"reflect"
+	"testing"
+)
+
+func getTestData(networkCIDR string, volumeProviderID string) *EtcdManagerOptions {
+	var o EtcdManagerOptions
+	o.InitDefaults()
+
+	o.ClusterName = "test"
+	o.BackupStorePath = "s3://test"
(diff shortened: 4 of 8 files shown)


ederst:
@justinsb, @hakman, @olemarkus anyone?

etcd-manager/cmd/etcd-manager/main.go
```diff
@@ -161,6 +163,9 @@ type EtcdManagerOptions struct {
 
 	// EtcdManagerMetricsPort allows exposing statistics from etcd-manager
 	EtcdManagerMetricsPort int
+
+	// IPFilter allows filtering for IP addresses with CIDR notation
+	IPFilter string
```
hakman, line 168: Seems like a good idea to me, just not sure about the name of the option. `NetworkCIDR` or `ClusterCIDR` may be closer to what kOps uses.
Also, in some cases there could be additional CIDRs, so multiple ones should be accepted.

What do you think @justinsb?

ederst: I agree, the naming is definitely not optimal, I'm open for suggestions. :)

To allow multiple opts I can maybe do smthg like this: https://github.com/kubernetes/cloud-provider-openstack/pull/1946

Either comma separated, or allow multiple cli opts, and priortize by order ("first come first serve").

ederst: @hakman I have updated the name to `NetworkCIDR` and it is now possible to specify multiple, comma separated CIDRs

ederst:
(@hakman, @justinsb, @olemarkus: I still want this to be merged, or a definitive no :D

etcd-manager/cmd/etcd-manager/main.go
```diff
@@ -189,6 +196,32 @@ func (o *EtcdManagerOptions) InitDefaults() {
 	o.Insecure = false
 	o.EtcdInsecure = false
 	o.EtcdManagerMetricsPort = 0
+
+	o.NetworkCIDR = os.Getenv("ETCD_MANAGER_NETWORK_CIDR")
```
hakman, line 200: Not quite sure we want this.

etcd-manager/cmd/etcd-manager/main.go
```diff
@@ -100,6 +100,8 @@ func main() {
 	var volumeTags stringSliceFlag
 	flag.Var(&volumeTags, "volume-tag", "tag which volume is required to have")
 
+	flag.StringVar(&o.NetworkCIDR, "network-cidr", o.NetworkCIDR, "filtering for IP addresses by defining a comma separated list of CIDRs.")
```
hakman, line 103: Please validate that this is only set for OpenStack, we want to make sure other providers don't expect it to work by mistake.

```suggestion
	flag.StringVar(&o.NetworkCIDR, "network-cidr", o.NetworkCIDR, "filtering for IP addresses by network CIDR (OpenStack only)")
```

hakman (changes requested):

etcd-manager/cmd/etcd-manager/main.go
```diff
@@ -201,6 +234,11 @@ func RunEtcdManager(o *EtcdManagerOptions) error {
 		return fmt.Errorf("backup-store is required")
 	}
 
+	networkCIDRs, err := parseNetworkCIDR(o)
```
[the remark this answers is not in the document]
hakman: Maybe override the value with the env var here.

etcd-manager/cmd/etcd-manager/main.go
ederst: This is just to make it easier to configure it within kOps, because currently the API config only exposes `env`, but no way to add parameters:
* https://pkg.go.dev/k8s.io/kops/pkg/apis/kops#EtcdManagerSpec

But probably it's possible to do scrap the ENV var and I need to find another way to configure this by either add a possibility to modify the parameters in the kOps API or some other supported way (by means of cloud-init, etc.).

etcd-manager/cmd/etcd-manager/main.go
```diff
@@ -189,6 +196,32 @@ func (o *EtcdManagerOptions) InitDefaults() {
...
+}
+
+func parseNetworkCIDR(o *EtcdManagerOptions) ([]*net.IPNet, error) {
+	if o.NetworkCIDR == "" {
+		return nil, nil
+	}
+
+	if o.VolumeProviderID != "openstack" {
+		return nil, fmt.Errorf("is only supported with provider 'openstack'")
+	}
```
ederst, line 210: @hakman you mentioned (in a now gone comment due to me committing the suggestion) to validate that it is only set for openstack, this should be enough?

I also test this here: https://github.com/kubernetes-sigs/etcdadm/pull/317/files#diff-6200c824e51c034077f4aaf4af4c567779591745d3f34ac7e0c53f55375a45e5R55

etcd-manager/cmd/etcd-manager/main.go
ederst: https://github.com/kubernetes-sigs/etcdadm/pull/317#discussion_r1122874600

etcd-manager/cmd/etcd-manager/main.go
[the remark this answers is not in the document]
ederst: > Maybe override the value with the env var here.

so if i understand you correctly, i should not do the [init with the env in `initDefaults()`](https://github.com/kubernetes-sigs/etcdadm/pull/317/files#r1122832991) but use the env var here?

but i still think the cli parameter should take precedence over the env var, as this is the common behavior afaik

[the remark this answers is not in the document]
hakman: Can you live without the env var until 1.27 is released and this is passed as a flag?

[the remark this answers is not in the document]
ederst: I think I can, so going to get rid of it

etcd-manager/cmd/etcd-manager/main.go
hakman: I think this function should be inlined. no need to complicate things.

etcd-manager/cmd/etcd-manager/main.go
```diff
@@ -202,6 +220,14 @@ func RunEtcdManager(o *EtcdManagerOptions) error {
 		return fmt.Errorf("backup-store is required")
 	}
 
+	if o.NetworkCIDR == "" {
+		o.NetworkCIDR = os.Getenv("ETCD_MANAGER_NETWORK_CIDR")
+	}
```
hakman, line 225: Let's remove the env var for now completely.

etcd-manager/pkg/volumes/openstack/util.go
```diff
@@ -26,20 +27,37 @@ const (
...
+	var fixedIPs []string
 	for _, address := range addrs {
 		if addresses, ok := address.([]interface{}); ok {
 			for _, addr := range addresses {
 				addrMap := addr.(map[string]interface{})
 				if addrType, ok := addrMap[openstackExternalIPType]; ok && addrType == openstackAddressFixed {
 					if fixedIP, ok := addrMap[openstackAddress]; ok {
 						if fixedIPStr, ok := fixedIP.(string); ok {
-							return fixedIPStr, nil
+							fixedIPs = append(fixedIPs, fixedIPStr)
```
hakman, line 39: Should return at the first IP that matches the subnet, when set.

etcd-manager/cmd/etcd-manager/main.go
ederst: Whoops, I thought I removed it.

etcd-manager/pkg/volumes/openstack/util.go
```diff
@@ -26,20 +27,37 @@ const (
...
+	if networkCIDR != nil {
+		for _, fixedIP := range fixedIPs {
+			if networkCIDR.Contains(net.ParseIP(fixedIP)) {
+				return fixedIP, nil
+			}
+		}
+	} else if len(fixedIPs) > 0 {
+		return fixedIPs[0], nil
+	}
+
```
ederst, line 61:
```suggestion
func GetServerFixedIP(addrs map[string]interface{}, name string, networkCIDR *net.IPNet) (poolAddress string, err error) {
	for _, address := range addrs {
		if addresses, ok := address.([]interface{}); ok {
			for _, addr := range addresses {
				addrMap := addr.(map[string]interface{})
				if addrType, ok := addrMap[openstackExternalIPType]; ok && addrType == openstackAddressFixed {
					if fixedIP, ok := addrMap[openstackAddress]; ok {
						if fixedIPStr, ok := fixedIP.(string); ok {
							if networkCIDR != nil && networkCIDR.Contains(net.ParseIP(fixedIP)) {
							    return fixedIP, nil
							} else {
							    return fixedIPs[0], nil
							}
						}
					}
				}
			}
		}
	}
```

etcd-manager/pkg/volumes/openstack/util.go
ederst: I am not sure about that. I can move the code to [GetServerFixedIP()](https://github.com/kubernetes-sigs/etcdadm/pull/317/files#diff-73f39bca80e8581052a90e491069fa8331e144ab25f30cadd33a958aaf25e3dfR49) again, but then this would add another layer of nesting to an already heavily nested, and at least for me hard to decipher code. This would look something like this:

https://github.com/kubernetes-sigs/etcdadm/pull/317/files#r1135238691

Although moving this to a separate function which first collects all the addresses and then filters them is more inefficient, it at least makes it more clearer what the code does IMO. But probably there is a better way of doing it, which I am not aware of?

etcd-manager/pkg/volumes/openstack/util.go
```diff
@@ -26,15 +27,21 @@ const (
...
 					if fixedIP, ok := addrMap[openstackAddress]; ok {
 						if fixedIPStr, ok := fixedIP.(string); ok {
-							return fixedIPStr, nil
+							if networkCIDR != nil {
+								if networkCIDR.Contains(net.ParseIP(fixedIPStr)) {
+									return fixedIPStr, nil
+								}
+							} else {
+								return fixedIPStr, nil
+							}
```
hakman, line 44:
```suggestion
							if networkCIDR != nil && !networkCIDR.Contains(net.ParseIP(fixedIPStr)) {
									continue
							}
							return fixedIPStr, nil
```

etcd-manager/cmd/etcd-manager/main.go
```diff
@@ -202,6 +207,19 @@ func RunEtcdManager(o *EtcdManagerOptions) error {
...
+	if o.NetworkCIDR != "" {
+		if o.VolumeProviderID != "openstack" {
+			return fmt.Errorf("network-cidr is only supported with provider 'openstack'")
+		}
+
+		_, networkCIDR, err = net.ParseCIDR(o.NetworkCIDR)
+		if err != nil {
+			return fmt.Errorf("network-cidr %s", err)
+		}
+	}
```
hakman, line 221:
```suggestion
	var networkCIDR *net.IPNet
	if o.NetworkCIDR != "" {
		if o.VolumeProviderID != "openstack" {
			return fmt.Errorf("network-cidr is only supported with provider 'openstack'")
		}

		var err error
		_, networkCIDR, err = net.ParseCIDR(o.NetworkCIDR)
		if err != nil {
			return fmt.Errorf("parsing network-cidr: %w", err)
		}
	}
```

hakman (changes requested):

hakman (approved):

PR #328, opened 2022-07-21
Fix reattachment of volumes for Hetzner

hakman:
If `volume.LocalDevice` is set etcd-manager will consider that the volume is mounted to the current server:
https://github.com/kubernetes-sigs/etcdadm/blob/7b9d5aafe83d5f94e37f00111b29b61d1800004c/etcd-manager/pkg/volumes/mounter.go#L299-L302


justinsb:
Thanks @hakman 

/approve
/lgtm

hakman:
Thanks @justinsb! :)

PR #334, opened 2022-08-25
Use volume name tag to match the instance group for Hetzner

hakman:
Goes together with https://github.com/kubernetes/kops/pull/14181.

```sh
$ hcloud volume list -o columns=name,server -s name 
NAME                                       SERVER
etcd-1.etcd-events.test1.k8s.local         master-fsn1-1-572f678e19233740
etcd-1.etcd-main.test1.k8s.local           master-fsn1-1-572f678e19233740
etcd-2.etcd-events.test1.k8s.local         master-fsn1-2-35d166ac6dc945b1
etcd-2.etcd-main.test1.k8s.local           master-fsn1-2-35d166ac6dc945b1
etcd-3.etcd-events.test1.k8s.local         master-fsn1-3-7af95061338a68fd
etcd-3.etcd-main.test1.k8s.local           master-fsn1-3-7af95061338a68fd
```

/hold for feedback
/cc @justinsb @olemarkus

### DIFF
--- a/etcd-manager/pkg/volumes/hetzner/volumes.go
+++ b/etcd-manager/pkg/volumes/hetzner/volumes.go
@@ -38,8 +38,9 @@ const (
 // HetznerVolumes defines the Hetzner Cloud volume implementation.
 type HetznerVolumes struct {
 	clusterName string
-	matchTags   map[string]string
-	nameTag     string
+
+	matchPeerTags map[string]string
+	matchNameTags map[string]string
 
 	hcloudClient *hcloud.Client
 	server       *hcloud.Server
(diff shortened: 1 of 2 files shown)


etcd-manager/pkg/volumes/hetzner/volumes.go
```diff
@@ -230,6 +230,9 @@ func getMatchingVolumes(client *hcloud.Client, matchLabels map[string]string) ([
 			labels = append(labels, k+"="+v)
 		}
 	}
+	if nameLabel != "" {
```
olemarkus, line 233: Should this logic be inside or outside of the function? I think I would expect not to have the extra function argument, but rather add the nameLabel to matchLabels.

hakman: How about now?

olemarkus (approved):

PR #336, opened 2022-08-26
Add support for Scaleway in etcd-manager

Mia-Cross:
As we are currently working on adding support for Scaleway in kOps, we also need etcd-manager to have a Scaleway volume provider.


etcd-manager/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.19-buster
```
remyleone, line 1: Why is this file required?

etcd-manager/Makefile
```diff
@@ -41,12 +41,19 @@ goimports:
...
 .PHONY: build-etcd-manager-amd64 build-etcd-manager-arm64
 build-etcd-manager-amd64 build-etcd-manager-arm64: build-etcd-manager-%:
+	# go build -o etcd-manager cmd/etcd-manager
 	${BAZEL} build ${BAZEL_FLAGS} --platforms=@io_bazel_rules_go//go/toolchain:linux_$* //cmd/etcd-manager:etcd-manager
 
 .PHONY: push-etcd-manager
 push-etcd-manager:
-	${BAZEL} run ${BAZEL_FLAGS} --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //images:push-etcd-manager
-	${BAZEL} run ${BAZEL_FLAGS} --platforms=@io_bazel_rules_go//go/toolchain:linux_arm64 //images:push-etcd-manager
+	echo $(REGISTRY_NAME)
```
remyleone, line 49: Why is this change required?

etcd-manager/cmd/etcd-manager/main.go
```diff
@@ -106,6 +109,8 @@ func main() {
 
 	fmt.Printf("etcd-manager\n")
 
+	fmt.Printf("HELLO, Looking for tags [%v]", volumeTags)
```
remyleone, line 112: Remove and use breakpoint instead in goland

etcd-manager/pkg/volumes/scaleway/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+
+	for _, volume := range etcdVolumes {
+		if volume.Server == nil {
+			// Volume doesn't have a server attached yet
+			continue
+		}
+		serverID := volume.Server.ID
+
+		// TODO: Get all servers in a single requests to reduce the number of calls to the could API
+		// Hetzner Cloud API is limited to 3600 requests/hour (see: https://docs.hetzner.cloud/#rate-limiting)
```
remyleone, line 48: copy and paste error

etcd-manager/pkg/volumes/scaleway/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+			return nil, fmt.Errorf("failed to get the running server: %s", err)
+		}
+		klog.V(2).Infof("Found the running server: %q", server.Server.Name)
+
+		if server.Server.PrivateIP == nil || *server.Server.PrivateIP == "" {
+			return nil, fmt.Errorf("failed to find private IP of server %s: ", serverID)
+		}
+		serverPrivateIP := server.Server.PrivateIP
+
+		klog.V(2).Infof("Discovered volume %s(%d) attached to server %s(%d)", volume.Name, volume.ID, server.Server.Name, serverID)
```
remyleone, line 63: Could you specify whether it is a block or local volume ?

etcd-manager/pkg/volumes/scaleway/volumes.go
```diff
@@ -0,0 +1,248 @@
...
+	instanceAPI *instance.API
+}
+
+var _ volumes.Volumes = &ScwVolumes{}
+
+// NewScwVolumes returns a new Scaleway Cloud volume provider.
+func NewScwVolumes(clusterName string, volumeTags []string, nameTag string) (*ScwVolumes, error) {
+	scwClient, err := scw.NewClient(
+		// TODO(Mia-Cross): Fix import for kopsv
+		//scw.WithUserAgent("kubernetes-kops/"+kopsv.Version),
```
remyleone, line 53: Is the User Agent handled somewhere else? It seems weird to handle it in a vendor-specific code.

etcd-manager/pkg/volumes/scaleway/volumes.go
```diff
@@ -0,0 +1,248 @@
...
+		// Only volumes from the same location can be mounted
+		if volume.Zone == "" {
+			klog.Warningf("failed to find volume location for %s(%d)", volume.Name, volume.ID)
+			continue
+		}
+		volumeLocation := volume.Zone
+		if volumeLocation != a.zone {
+			continue
+		}
+		klog.V(2).Infof("Found attachable volume %s(%d) with status %q", volume.Name, volume.ID, volume.State)
```
remyleone, line 130: Could you also indicate the type of volume?

etcd-manager/pkg/volumes/scaleway/volumes.go
```diff
@@ -0,0 +1,248 @@
...
+		//err = a.instanceAPI.ServerActionAndWait(&instance.ServerActionAndWaitRequest{
+		//	ServerID: a.server.ID,
+		//	Zone:     a.zone,
+		//	Action:   instance.ServerActionPoweroff,
+		//})
+		//if err != nil {
+		//	return fmt.Errorf("failed to power-off server %s before attaching volume", a.server.ID)
+		//}
+
+		klog.V(2).Infof("Attaching volume %s(%d) to the running server", scwVolume.Name, scwVolume.ID)
```
remyleone, line 203: Right now let's support only block volume that can be added without restarting a server. That seems easier to handle for a first version rather than having local volume also supported.

etcd-manager/pkg/volumes/scaleway/volumes.go
```diff
@@ -0,0 +1,248 @@
...
+		//_, err = a.instanceAPI.ServerAction(&instance.ServerActionRequest{
+		//	ServerID: a.server.ID,
+		//	Zone:     a.zone,
+		//	Action:   instance.ServerActionPoweron,
+		//})
+		//if err != nil {
+		//	return fmt.Errorf("failed to power server %s back on after attaching volume", a.server.ID)
+		//}
+
+		time.Sleep(10 * time.Second)
```
remyleone, line 224: Use a wait operation on this. Having a static wait is not performant nor elegant.

etcd-manager/Dockerfile
Mia-Cross: It's only for dev purposes, it won't go in the final PR. I need it to be able to build an image containing both etcd-manager executable and sh since the command launching the etcd-manager image is generated by kops (hard to modify) and is in the form of `/bin/sh -c exec /etcd-manager [arguments]`

etcd-manager/Makefile
Mia-Cross: The echo is not required indeed and was only used for debugging a script I was making which purpose was to build and push to the registry all the images needed by kOps across several repositories.
If you're talking about why I got rid of the Bazel commands it's because they required me to write BUILD.bazel files for the sdk and it seems like a lot of work that I was not sure was going to pay at the time. But now that I know that it works, I will eventually need to write these Bazel files.

etcd-manager/Dockerfile
remyleone: Maybe it could be contributed in a different PR to etcdmanager? Maybe other people would benefit from it.

etcd-manager/Makefile
```diff
@@ -20,6 +20,13 @@ IMAGE_BASE                 := $(STABLE_DOCKER_REGISTRY)/$(STABLE_DOCKER_IMAGE_PR
 BAZEL?=bazelisk
 BAZEL_FLAGS=--features=pure --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64
 
+#PASSWORD=$(shell cat /home/leila/.scw/hashicorp/skey)
```
remyleone, line 23: I would suggest migrating that in the configuration of your run target in the IDE

etcd-manager/Makefile
hakman: @Mia-Cross The `The BUILD.bazel` files are auto-generated by running `make vendor`.

etcd-manager/pkg/volumes/scaleway/volumes.go
```diff
@@ -0,0 +1,262 @@
...
+
+	//metadataAPI := instance.NewMetadataAPI()
+	//metadata, err := metadataAPI.GetMetadata()
+	//if err != nil {
+	//	return nil, fmt.Errorf("failed to retrieve server metadata: %s", err)
+	//}
+	//TODO(Mia-Cross): The previous segment only works if launched from instance
+	// For developpement purposes, we can execute the following code instead
+	master, err := instance.NewAPI(scwClient).ListServers(&instance.ListServersRequest{
+		Zone: scw.Zone(os.Getenv("SCW_DEFAULT_ZONE")),
```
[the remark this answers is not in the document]
hakman: Could be passed by kOps as global env var, same way it is done for other clouds in [buildEnvironmentVariables()](
https://github.com/kubernetes/kops/blob/921eadd04f16af8cd2f686942177c441679b6f98/pkg/model/bootstrapscript.go#L131).

etcd-manager/pkg/volumes/scaleway/discovery.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"fmt"
+
+	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/etcdadm/etcd-manager/pkg/privateapi/discovery"
+)
+
+// ScwVolumes also allows the discovery of peer nodes
```
justinsb, line 27: Go style suggests just calling this `Volumes`, it will be named `scaleway.Volumes` when imported outside of the package.  See "Avoid repetition" https://go.dev/blog/package-names

Just a nit / FYI though - not a blocker for merge.

etcd-manager/pkg/volumes/scaleway/discovery.go
```diff
@@ -0,0 +1,71 @@
...
+			continue
+		}
+		serverID := volume.Server.ID
+
+		server, err := a.instanceAPI.GetServer(&instance.GetServerRequest{
+			ServerID: serverID,
+			Zone:     a.zone,
+		})
+		if err != nil || server == nil {
+			return nil, fmt.Errorf("failed to get the running server: %s", err)
```
justinsb, line 52: Nit: when wrapping errors (which is a good idea), we should use `: %w` now (in the past %v was recommended, but %w preserves the "inner" error in a form that can be inspected from code)

etcd-manager/pkg/volumes/scaleway/volumes.go
```diff
@@ -0,0 +1,226 @@
...
+
+var _ volumes.Volumes = &ScwVolumes{}
+
+// NewScwVolumes returns a new Scaleway Cloud volume provider.
+func NewScwVolumes(clusterName string, volumeTags []string, nameTag string) (*ScwVolumes, error) {
+	scwClient, err := scw.NewClient(
+		scw.WithEnv(),
+	)
+	if err != nil {
+		return nil, err
```
justinsb, line 53: Nit: I generally suggest wrapping errors when they come from code outside of "our" projects (of course you might consider the scaleway SDK one of "our" projects, and if the errors are self-descriptive, then it's fine _not_ to wrap).  The AWS SDK errors were/are pretty terse, so without wrapping they can be hard to decipher.

etcd-manager/pkg/volumes/scaleway/volumes.go
```diff
@@ -0,0 +1,226 @@
...
+		if err != nil || volumeResp.Volume == nil {
+			return fmt.Errorf("failed to get info for volume id %q: %w", volume.ProviderID, err)
+		}
+
+		scwVolume := volumeResp.Volume
+		if scwVolume.Server != nil {
+			if scwVolume.Server.ID != a.server.ID {
+				return fmt.Errorf("found volume %s(%s) attached to a different server: %s", scwVolume.Name, scwVolume.ID, scwVolume.Server.ID)
+			}
+			klog.V(2).Infof("Attached volume %s(%d) of type %s to the running server", scwVolume.Name, scwVolume.ID, scwVolume.VolumeType)
```
justinsb, line 174: Nit: Maybe `volume %s(%d) of type %s is already attached to the running server`

justinsb:
I did a pass over the go code and it looks good - a few style points, but more just FYI, nothing blocking.

/ok-to-test

olemarkus (approved):
/lgtm

PR #341, opened 2022-09-02
Update dependencies before adding support for Scaleway

Mia-Cross:
Before we can add support for Scaleway in etcd-manager (#336), we first need to update the modules, particularly kops, so that we can incorporate the changes made in [this PR](https://github.com/kubernetes/kops/pull/14214) in the vendor directory.
The update resulted in an error with the mount.Interface so a few changes were necessary in etcd-manager/pkg/hostmount/nsenter.go


olemarkus (approved):
/lgtm

PR #344, opened 2022-10-26
Support for scaleway without multi-attach of volumes

Mia-Cross:
Scaleway doesn't allow multi-attach of volumes, that's' why we had to make a few changes in order for the etcd-manager to work with Scaleway.


Mia-Cross:
This PR was made in order to support master instance groups with more than one instance, which I recently learned we're not really supposed to do so it's not useful. The new one is #345.

PR #345, opened 2022-11-30
Scaleway support in etcd-manager operational

Mia-Cross:
My previous PR (#344) was made in order to support master instance groups with more than one instance, which I recently learned we're not really supposed to do. This new one only fixes a typo in a print and adds a few comments to the code but with it the support for Scaleway in etcd-manager is complete.


Mia-Cross:
I fixed the problem by recovering a part of my old work in multi-masters : the way I do it is by adding the instance-group tag to the volume in kops. Then in etcd-manager I can look for volumes to attach with this tag and remove it when I look for volumes in Poll()
Is it ok ?

hakman:
I think the more correct approach would be to send that tag as `VolumeNameTag` instead.
This way you can use `VolumeTag` for finding peers and and add `VolumeNameTag` when looking for the exact volume to mount.
```go
config.VolumeTag = []string{
	fmt.Sprintf("%s=%s", scaleway.TagClusterName, b.Cluster.Name),
	fmt.Sprintf("%s=%s", scaleway.TagNameEtcdClusterPrefix, etcdCluster.Name),
}
config.VolumeNameTag = fmt.Sprintf("%s=%s", scaleway.TagInstanceGroup, instanceGroupName)
```
vs
```go
config.VolumeTag = []string{
	fmt.Sprintf("%s=%s", scaleway.TagClusterName, b.Cluster.Name),
	fmt.Sprintf("%s=%s", scaleway.TagNameEtcdClusterPrefix, etcdCluster.Name),
	fmt.Sprintf("%s=%s", scaleway.TagNameRolePrefix, scaleway.TagRoleMaster),
	fmt.Sprintf("%s=%s", scaleway.TagInstanceGroup, instanceGroupName),
}
config.VolumeNameTag = fmt.Sprintf("%s=%s", scaleway.TagNameEtcdClusterPrefix, etcdCluster.Name)
```

Mia-Cross:
I had a lot of racing problems if the name tag was only the instance group name so I added the name of the etcd cluster to make sure that the `FindVolumes()` function only returns one volume to attach :
```
config.VolumeTag = []string{
				fmt.Sprintf("%s=%s", scaleway.TagClusterName, b.Cluster.Name),
				fmt.Sprintf("%s=%s", scaleway.TagNameEtcdClusterPrefix, etcdCluster.Name),
				fmt.Sprintf("%s=%s", scaleway.TagNameRolePrefix, scaleway.TagRoleMaster),
			}
			config.VolumeNameTag = fmt.Sprintf("%s=%s-%s", scaleway.TagInstanceGroup, instanceGroupName, etcdCluster.Name)
```

Mia-Cross:
@hakman I've made new changes according to your comments on [my kops PR](https://github.com/kubernetes/kops/pull/14641), it should be good now.

hakman:
Nice. Thanks @Mia-Cross. Happy with the end result?

Mia-Cross:
Yes ! Thanks for your advice :)

PR #348, opened 2022-12-06
Restart the server using the recommended etcd version

hakman:
When re-starting after an upgrade, `etcd-manager` tries to use the version of `etcd` from the state. This makes it impossible for us to remove any unused `etcd` versions from the bundle, even if there is a newer patch version that should be used instead.

Same approach as when restoring from backup:
https://github.com/kubernetes-sigs/etcdadm/blob/f9a20374ca1abc9a6083e6db21aeaa32e2560504/etcd-manager/pkg/etcd/restore.go#L125-L138

/cc @justinsb @olemarkus

### DIFF
--- a/etcd-manager/pkg/etcd/etcdserver.go
+++ b/etcd-manager/pkg/etcd/etcdserver.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/etcdadm/etcd-manager/pkg/backup"
 	"sigs.k8s.io/etcdadm/etcd-manager/pkg/contextutil"
 	"sigs.k8s.io/etcdadm/etcd-manager/pkg/dns"
+	"sigs.k8s.io/etcdadm/etcd-manager/pkg/etcdversions"
 	"sigs.k8s.io/etcdadm/etcd-manager/pkg/legacy"
 	"sigs.k8s.io/etcdadm/etcd-manager/pkg/pki"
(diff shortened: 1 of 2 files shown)


hakman:
@justinsb Any chance you could take a look at this PR?

justinsb:
Q: Do we need this if we're going to get the exact etcd version via a mount/initContainer/etc?

hakman:
> Q: Do we need this if we're going to get the exact etcd version via a mount/initContainer/etc?

Yes. We still need this for upgrades, in case the cluster was previously using a version that we don't recommend anymore.

etcd-manager/pkg/etcd/etcdserver.go
```diff
@@ -604,12 +605,22 @@ func (s *EtcdServer) startEtcdProcess(state *protoetcd.EtcdState) error {
 		return err
 	}
 
-	binDir, err := BindirForEtcdVersion(state.EtcdVersion, "etcd")
```
justinsb, line 607: So I'm sorry, but I'm still worried about this... 

This particular method (startEtcdProcess) is called from multiple places, and I think for some of them we should use `EtcdVersionForAdoption`, and for some of them we should use the version the caller asked for.

I also think we should report our version accurately in calls to GetInfoRequest, so we should populate state.EtcdVersion with whatever version we are going to run.

I'm digging into the details of these callsites though - you're usually right on these things, but I need to make sure :-)

PR #349, opened 2022-12-06
Winter code cleanup

hakman:
Some issues still remain, but very few:
```
> staticcheck -- ./... | grep -v -e "\.pb\.go" -e "proto" 
pkg/controller/controller.go:779:28: this result of append is never used, except maybe in other appends (SA4010)
pkg/etcdclient/client.go:144:2: empty branch (SA9003)
pkg/hostmount/nsenter.go:47:15: ListProcMounts not declared by package mount (compile)
pkg/hostmount/nsenter.go:94:21: MakeMountArgs not declared by package mount (compile)
pkg/hostmount/nsenter.go:112:31: AddSystemdScope not declared by package mount (compile)
pkg/pki/fs_test.go:126:28: actual.CertPool().Subjects has been deprecated since Go 1.18: if s was returned by SystemCertPool, Subjects will not include the system roots.  (SA1019)
pkg/privateapi/peers.go:301:23: grpc.WithInsecure is deprecated: use WithTransportCredentials and insecure.NewCredentials() instead. Will be supported throughout 1.x.  (SA1019)
pkg/privateapi/peers.go:304:22: grpc.WithBackoffMaxDelay is deprecated: use WithConnectParams instead. Will be supported throughout 1.x.  (SA1019)
```

/cc @olemarkus @justinsb


etcd-manager/pkg/controller/controller.go
```diff
@@ -950,13 +950,13 @@ func (m *EtcdController) removeNodeFromCluster(ctx context.Context, clusterSpec
 					peerState := m.peerState[privateapi.PeerId(id)]
 					if peerState == nil {
 						klog.Fatalf("peerState unexpectedly nil")
+					} else {
```
johngmyers, line 953: This change seems less desirable than what was before.

hakman: More or less agree, but `staticheck` was seeing a possible issue with the code. It's not so bad as it is now either.

hakman: Any objection to merging this, now that this has been reverted?

PR #356, opened 2023-01-13
Update etcdadm build

hakman:
/cc @justinsb @dlipovetsky


cmd/version.go
```diff
@@ -22,8 +22,8 @@ import (
 
 	log "sigs.k8s.io/etcdadm/pkg/logrus"
 
-	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
```
justinsb, line 26: I agree with this one, but technically this might be a breaking change, because it changes the recognized tags from `yaml` to `json`.  Can we split this one into a separate PR?

justinsb:
Thanks for fixing - all looks good except can we split that yaml change into a separate PR - do we need to do it here?  I think we also have to move tags from `yaml` to `json` to do it.

cmd/version.go
justinsb: Ooops, you corrected me in office hours, I was confused :-)

hakman:
> Thanks for fixing - all looks good except can we split that yaml change into a separate PR - do we need to do it here? I think we also have to move tags from `yaml` to `json` to do it.

Addressed, as noted in https://github.com/kubernetes-sigs/etcdadm/pull/356#discussion_r1081374788.

hakman:
I think this can be merged and fix `cluster-create-phases-test` separately.

justinsb (approved):

justinsb:
/override cluster-create-phases-test

Unrelated, and I'm looking at fixing it separately

hakman:
> Unrelated, and I'm looking at fixing it separately

@justinsb I commented the failing job to allow this to merge.

justinsb:
/approve
/lgtm

I think the failing job is a regression in the 3.4 client library.  My plan is: go back to the 3.4 client libraries, re-enable the test, hopefully see it passing, create a PR with the 3.5 client libraries, show it is failing, open upstream.  I think we're probably OK with the 3.4 client libs for now, it would just be good to not be stuck on them.  Also I think it's a real regression that has consequences for kubernetes, so probably important to fix.  But not in this PR :-)

PR #361, opened 2023-02-03
Restart the server using the recommended etcd version

justinsb:
When we load a version from disk, we run the latest recommended patch
version.  This enables us to depend on fewer versions of etcd.
    
 Co-authored-by: Justin SB


justinsb:
I think this is a minimal version - WDYT @hakman ?

/assign @hakman 

Of course, the really tricky thing is that we need to figure out how to test this, we need an upgrade test that runs an otherwise unsupported version of etcd I guess and then does an upgrade...

hakman:
Thanks @justinsb. Looks great!
We have in kOps a test for upgrade which should be good in this case: `pull-kops-e2e-aws-upgrade-126-ko126-to-klatest-kolatest-many-addons`.

/lgtm

PR #363, opened 2023-02-04
Use etcd 3.4 client

justinsb:
3.5 seems to have regressed on https://github.com/kubernetes/kubernetes/issues/72102


justinsb:
This is looking good I think ... 3.4 is passing, 3.5 is failing only on the test which shows the regression.  Going to merge this one.

/hold cancel

PR #365, opened 2023-02-04
Update to go 1.19.5

justinsb:
Also requires a bazel update.


etcd-manager/.bazelversion
```diff
@@ -1 +1 @@
-4.2.1
+5.4.0
```
hakman, line 1: Bazel requires new images for cloudbuild, which didn't exist until recently:
https://github.com/kubernetes-sigs/etcdadm/blob/master/etcd-manager/cloudbuild.yaml

See: https://github.com/bazelbuild/continuous-integration/issues/1060#issuecomment-1369119958

justinsb: Ouch - thanks - I'll take a look.

It would be great to get off bazel - it's kinda silly that to build with the newer go we need to update the tasks, and they require updating bazel, and that requires a cloudbuild image.  Just too many pieces in the chain there!

hakman: I agree. For now, #366 should fix the promotion.

PR #369, opened 2023-04-01
Switched from N1 to E2 machine

sarab97:
Reference https://github.com/kubernetes/k8s.io/issues/5059

This changes machine type from N1 to E2 which are newer but almost same pricing. But with E2 being more efficient and overall taking less time. The cost can be reduced.


justinsb:
Thanks @sarab97

/approve
/lgtm

PR #296, opened 2022-02-16
`make container-build` support GOPROXY env

FOWind:
Related issue: #295 

1. Modify makefile to support GOPROXY on `container-build` command
2. Add `GOPROXY` use case in README

PR #387, opened 2023-12-24
Specify that using GOPROXY is not necessary for all users and remove implied endorsement of goproxy.io

abctaylor:
PR for Issue #296


justinsb:
Thanks @abctaylor , this makes it much clearer when to use GOPROXY and when not to.

Are you able to sign the CLA?

hakman (approved):

PR #389, opened 2024-03-29
Update dependencies

hakman:


etcd-manager/images/BUILD
```diff
@@ -92,10 +92,6 @@ container_image(
 
 container_image(
     name = "etcd-manager",
-    architecture = select({
-        "@io_bazel_rules_go//go/platform:amd64": "amd64",
-        "@io_bazel_rules_go//go/platform:arm64": "arm64",
-    }),
```
rifelpet, line 98: why are these being removed? have we tested that the resulting images are multi-arch?

perhaps we should follow the archived rules_docker suggestion and [move to rules_oci](https://github.com/bazelbuild/rules_docker?tab=readme-ov-file#status)

hakman: I am just experimenting. Bazel anything is huge pain. Tests are not yet passing due to build errors.
Not sure I want to move to rules_oci, just to strip it away in a few months.

hakman: Went back to the basics and it's good to go now.
@rifelpet Could you take another look?

hakman: BTW, for some reason this did not compile with Go 1.22...

rifelpet (approved):

PR #396, opened 2024-04-27
Fix the issue that an incorrect member may be removed

ahrtr:
The second parameter "member" is shadowned by the variable "member" in the for loop. So it may remove an incorrect member.

cc @hakman @justinsb


hakman:
Nice catch, thanks!
/ok-to-test

hakman (approved):

